PR #18, opened 2025-07-19
feat: add authorization token ml-9

MrKopchick:
Backend:
added a jwt service in src/libs/modules/token

Frontend:
added a jwt helper in src/libs/helpers


apps/backend/src/libs/modules/token/jwt.service.ts
```diff
@@ -0,0 +1,24 @@
+import { BaseToken } from "./base-token.js";
+import { type JwtPayload } from "./libs/types/types.js";
+
+const SECRET = new TextEncoder().encode(process.env["JWT_SECRET"]);
+const EXPIRES_IN = process.env["JWT_EXPIRES_IN"] ?? "24h";
```
fshabanov, line 5: Use config module instead. Check how we use other env variables

fshabanov (changes requested):

apps/backend/src/libs/modules/token/base-token.ts
```diff
@@ -0,0 +1,41 @@
...
+	private algorithm: string;
+	private secret: Uint8Array;
+
+	constructor({ algorithm, secret }: Constructor) {
+		this.secret = secret;
+		this.algorithm = algorithm;
+	}
+
+	public async createToken({
+		expirationTime,
```
fshabanov, line 23: Let's add expirationTime through the constructor since it comes from .env

apps/backend/src/libs/modules/token/jwt.service.ts
```diff
@@ -0,0 +1,26 @@
+import { type BaseToken } from "./base-token.js";
+import { type JwtPayload } from "./libs/types/types.js";
+
+class JwtService {
```
fshabanov, line 4: Why do we need this service at all? It does the same thing as BaseToken. We can only have BaseToken and that will be enough

apps/backend/src/libs/modules/token/base-token.ts
```diff
@@ -0,0 +1,41 @@
+import {
```
fshabanov, line 1: Rename the file to `base-token.module.ts`

apps/backend/src/libs/modules/token/jwt.ts
```diff
@@ -0,0 +1,18 @@
...
+import { type JwtPayload } from "./libs/types/types.js";
+
+const SECRET = new TextEncoder().encode(config.ENV.TOKEN.SECRET);
+const ALGORITHM = config.ENV.TOKEN.ALGORITHM;
+const EXPIRES_IN = config.ENV.TOKEN.EXPIRES_IN;
+
+const baseToken = new BaseToken<JwtPayload>({
+	algorithm: ALGORITHM,
+	secret: SECRET,
+});
```
fshabanov, line 14:
```suggestion

const baseToken = new BaseToken<JwtPayload>({
	algorithm: config.ENV.TOKEN.ALGORITHM,
	secret: new TextEncoder().encode(config.ENV.TOKEN.SECRET),
});
```

No need to create new variables here

fshabanov (changes requested):

apps/backend/src/libs/modules/token/base-token.module.ts
```diff
@@ -0,0 +1,38 @@
...
+		this.expirationTime = expirationTime;
+	}
+
+	public async sign(payload: T): Promise<string> {
+		return await new SignJWT(payload)
+			.setProtectedHeader({ alg: this.algorithm })
+			.setIssuedAt()
+			.setExpirationTime(this.expirationTime)
+			.sign(this.secret);
+	}
```
o-nedashkivska, line 31: Seems like no async await needed here. The same for verify method
```suggestion
	public sign(payload: T): Promise<string> {
		return new SignJWT(payload)
			.setProtectedHeader({ alg: this.algorithm })
			.setIssuedAt()
			.setExpirationTime(this.expirationTime)
			.sign(this.secret);
	}
```

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -4,17 +4,23 @@ import {
 } from "~/modules/users/libs/types/types.js";
 import { type UserService } from "~/modules/users/user.service.js";
 
+import { jwt } from "../../libs/modules/token/jwt.js";
```
fshabanov, line 7:
```suggestion
import { jwt } from "~/libs/modules/token/jwt.js";
```

apps/backend/src/libs/modules/token/jwt.ts
```diff
@@ -0,0 +1,12 @@
+import { config } from "~/libs/modules/config/config.js";
```
fshabanov, line 1: Rename the file name to `token.ts` since the module folder is named `token`

fshabanov (changes requested):

fshabanov:
Pull the changes from main branch

apps/backend/src/libs/modules/token/token.ts
```diff
@@ -0,0 +1,12 @@
+import { config } from "~/libs/modules/config/config.js";
+
+import { BaseToken } from "./base-token.module.js";
+import { type JwtPayload } from "./libs/types/types.js";
+
+const jwt = new BaseToken<JwtPayload>({
+	algorithm: config.ENV.TOKEN.ALGORITHM,
+	expirationTime: config.ENV.TOKEN.EXPIRES_IN,
+	secret: new TextEncoder().encode(config.ENV.TOKEN.SECRET),
+});
```
sofiia-trokhymchuk, line 10:
```suggestion
const { ALGORITHM, EXPIRES_IN, SECRET } = config.ENV.TOKEN;

const jwt = new BaseToken<JwtPayload>({
	algorithm: ALGORITHM,
	expirationTime: EXPIRES_IN,
	secret: new TextEncoder().encode(SECRET),
});
```

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -72,8 +72,13 @@ class AuthController extends BaseController {
 			body: UserSignUpRequestDto;
 		}>,
 	): Promise<APIHandlerResponse> {
+		const { token, user } = await this.authService.signUp(options.body);
+
 		return {
-			payload: await this.authService.signUp(options.body),
+			payload: {
+				token,
```
fshabanov, line 79: Save this token in frontend, since we already have signUp action there. Use storage module that is already created, do not create a new helper

sofiia-trokhymchuk (changes requested):

sofiia-trokhymchuk (approved):

packages/shared/src/modules/users/libs/types/user-response-dto.type.ts
```diff
@@ -0,0 +1,9 @@
+type UserResponseDto = {
+	email: string;
+	firstName?: string;
+	id: number;
+	lastName?: string;
+	photoSrc?: string;
```
fshabanov, line 6: We don't have photoSrc right now, remove it

fshabanov (changes requested):

fshabanov (approved):

PR #19, opened 2025-07-19
feat: add get authenticated user ml-15

VolodymyrKozel:
### Summary
- Added `getAuthenticatedUser` method in  `AuthController`
- Create type AuthHandlerOptions to include `user`
- Added OpenAPI Swagger docs for authenticated user endpoint


apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -37,6 +43,35 @@ class AuthController extends BaseController {
...
+	 *                properties:
+	 *                  message:
+	 *                    type: object
+	 *                    $ref: "#/components/schemas/User"
+	 */
+
+	private getAuthenticatedUser(
+		options: APIHandlerOptions,
+	): Promise<APIHandlerResponse> {
+		const { user } = options;
```
fshabanov, line 67: But we don't have user in options, where does it come from?

fshabanov (changes requested):

apps/backend/src/libs/modules/controller/libs/types/api-handler-options.type.ts
```diff
@@ -4,12 +4,14 @@ type APIHandlerOptions<
 	body: T["body"];
 	params: T["params"];
 	query: T["query"];
+	user: T["user"];
```
fshabanov, line 7: It's not correct to add user in this type, because we will not always have a user. For white routes there will be no user. You can use type casting in auth controller instead

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -37,6 +43,35 @@ class AuthController extends BaseController {
 		});
 	}
 
+	/**
+	 * @swagger
+	 * /auth/authenticated-user:
+	 *    get:
+	 *      description: Get authenticated user's data
+	 *      responses:
+	 *        200:
```
fshabanov, line 52: Don't forget to add error responses as well

apps/backend/src/modules/auth/auth.controller.ts
fshabanov: Got it. In that case we will need to wait for Serhii to finalize checking your PR

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -37,6 +43,35 @@ class AuthController extends BaseController {
...
+	private getAuthenticatedUser(
+		options: APIHandlerOptions,
+	): Promise<APIHandlerResponse> {
+		const { user } = options;
+
+		return Promise.resolve({
+			payload: user,
+			status: HTTPCode.OK,
+		});
+	}
```
fshabanov, line 73:
```suggestion
		return {
			payload: user ?? null,
			status: HTTPCode.OK,
		};
	}
```

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -11,6 +12,10 @@ class AuthService {
 		this.userService = userService;
 	}
 
+	public async getAuthenticatedUser(userId: number): Promise<UserResponseDto> {
+		return await this.userService.find(userId);
+	}
+
```
fshabanov, line 18: You don't seem to use this anywhere

packages/shared/src/modules/users/libs/types/user-response-dto.type.ts
```diff
@@ -0,0 +1,6 @@
+type UserResponseDto = {
+	email: string;
+	id: number;
+};
```
fshabanov, line 4: User also has createdAt, updatedAt fields

fshabanov (changes requested):

apps/backend/src/modules/auth/auth.controller.ts
fshabanov: Makes sense. But we still do not need Promise.resolve here, you can just return an object

apps/backend/src/modules/auth/libs/types/auth-handler-options.type.ts
```diff
@@ -0,0 +1,8 @@
+import { type APIHandlerOptions } from "~/libs/modules/controller/controller.js";
+import { type UserResponseDto } from "~/modules/users/libs/types/types.js";
+
+type AuthHandlerOptions = APIHandlerOptions & {
+	user: UserResponseDto;
+};
```
fshabanov, line 6: This type will be used only in one, I think we don't need to create a separate type for it. You can directly use `as APIHandlerOptions<{ user: UserResponseDto; }>`

fshabanov (changes requested):

apps/backend/src/libs/modules/controller/libs/types/api-handler-options.type.ts
```diff
@@ -1,11 +1,11 @@
 type APIHandlerOptions<
-	T extends DefaultApiHandlerOptions = DefaultApiHandlerOptions,
-> = {
+	T extends Partial<DefaultApiHandlerOptions> &
+		Record<string, unknown> = DefaultApiHandlerOptions,
+> = Omit<T, keyof DefaultApiHandlerOptions> & {
```
fshabanov, line 4: Sorry, my bad, we do need to add `user` actually instead of this 😅

fshabanov (approved):
We can merge after Serhii's PR is merged

apps/backend/src/libs/modules/controller/libs/types/api-handler-options.type.ts
VolodymyrKozel: I, revert last commit.

Issue #9, opened 2025-07-17
Authorization token (JWT)


o-nedashkivska:
Implement JWT token signing and verification

https://www.npmjs.com/package/jose

- [x] The token should be generated and returned by sign-in and sign-up endpoints
- [x] The token should store userId

PR #20, opened 2025-07-19
feat: add sign in ml-5

Karpt221:
- Added shared `signRequesrDto`, validation schema, api path and `AuthError`
- Added hashing
- Implemented `find()` and `findByEmail()` in `UserRepository` and `UserService`
- Added signIn in `AuthService` and `AuthController`


apps/backend/src/modules/users/libs/hash/hash.ts
```diff
@@ -0,0 +1,20 @@
+import * as bcrypt from "bcrypt";
+
+async function hashPassword(
```
fshabanov, line 3: You should create a separate bcrypt module in libs/modules

apps/backend/src/modules/users/libs/types/user-attribute.type.ts
```diff
@@ -0,0 +1,7 @@
+import { ValueOf } from "@meetlytic/shared";
```
fshabanov, line 1:
```suggestion
import { type ValueOf } from "~/libs/types/types.js";
```

apps/backend/src/modules/users/libs/types/user-attribute.type.ts
```diff
@@ -0,0 +1,7 @@
...
+
+import { UserAttribute as UserAttributeEnum } from "../enums/enums.js";
```
fshabanov, line 3: B17

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -28,8 +32,15 @@ class UserRepository implements Repository {
...
 
-	public find(): ReturnType<Repository["find"]> {
-		return Promise.resolve(null);
+	public async find({
+		attribute,
+		value,
+	}: {
+		attribute: UserAttribute;
+		value: Objection.PrimitiveValue;
+	}): Promise<null | UserEntity> {
```
fshabanov, line 41: `find` method should only search by id. To search by other fields add new methods, such as `findByEmail`

apps/backend/src/modules/users/user.service.ts
```diff
@@ -33,8 +38,19 @@ class UserService implements Service {
 		return Promise.resolve(true);
 	}
 
-	public find(): ReturnType<Service["find"]> {
-		return Promise.resolve(null);
+	public async find({
+		attribute,
+		value,
+	}: {
```
fshabanov, line 44: Same about find method

fshabanov (changes requested):

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,16 +1,53 @@
...
+		);
+		if (!isValid) {
+			throw new HTTPError({
+				message: "Not valid password",
+				status: HTTPCode.UNAUTHORIZED,
+			});
+		}
+		return {
+			token: await jwt.sign({ userId: user.id }),
+			user: { email: user.email, id: user.id },
```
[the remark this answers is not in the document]
Karpt221: But if I use user directly from userService in return of signIn then I it will return password hash and salt to the client. At the same time, hash is required for password verification. So I think it would be better to filter user object keeping only email and id.

apps/backend/src/libs/modules/bcrypt/base-bcrypt.module.ts
```diff
@@ -0,0 +1,20 @@
+import * as bcrypt from "bcrypt";
```
o-nedashkivska, line 1:
```suggestion
import bcrypt from "bcrypt";
```

apps/backend/src/libs/modules/bcrypt/base-bcrypt.module.ts
```diff
@@ -0,0 +1,20 @@
...
+		return { hash, salt };
+	}
+
+	async verify(
+		plainPassword: string,
+		hashedPassword: string,
+	): Promise<boolean> {
+		const match = await bcrypt.compare(plainPassword, hashedPassword);
+		return match;
+	}
```
o-nedashkivska, line 17: There is no need to create an extra variable
```suggestion
	verify(
		plainPassword: string,
		hashedPassword: string,
	): Promise<boolean> {
		return bcrypt.compare(plainPassword, hashedPassword);
	}
```

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -35,6 +37,61 @@ class AuthController extends BaseController {
...
+	 *        200:
+	 *          description: Successful operation
+	 *          content:
+	 *            application/json:
+	 *              schema:
+	 *                type: object
+	 *                properties:
+	 *                  message:
+	 *                    type: object
+	 *                    $ref: "#/components/schemas/User"
```
o-nedashkivska, line 84: You should also describe errors you may throw if something went wrong

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -28,8 +30,9 @@ class UserRepository implements Repository {
 		return Promise.resolve(true);
 	}
 
-	public find(): ReturnType<Repository["find"]> {
-		return Promise.resolve(null);
+	public async find(id: number): Promise<null | UserEntity> {
+		const [user] = await this.userModel.query().where(UserAttribute.ID, id);
```
o-nedashkivska, line 34: Try using findById here https://vincit.github.io/objection.js/api/query-builder/find-methods.html#findbyid

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -38,6 +41,13 @@ class UserRepository implements Repository {
 		return users.map((user) => UserEntity.initialize(user));
 	}
 
+	public async findByEmail(email: string): Promise<null | UserEntity> {
+		const [user] = await this.userModel
+			.query()
+			.where(UserAttribute.EMAIL, email);
```
o-nedashkivska, line 47: try using findByOne here https://vincit.github.io/objection.js/api/query-builder/find-methods.html#findbyid

packages/shared/src/modules/users/libs/validation-schemas/user-sign-in.validation-schema.ts
```diff
@@ -0,0 +1,25 @@
...
+		email: z
+			.string()
+			.trim()
+			.min(UserValidationRule.EMAIL_MINIMUM_LENGTH, {
+				message: UserValidationMessage.EMAIL_REQUIRE,
+			})
+			.email({
+				message: UserValidationMessage.EMAIL_WRONG,
+			}),
+		password: z.string().trim(),
```
o-nedashkivska, line 21: Seems like we may need to have some minimum length here as well

apps/backend/src/modules/users/user.service.ts
```diff
@@ -22,7 +22,7 @@ class UserService implements Service {
 	public async create(
```
Karpt221, line 19: The payload of the user service create method will be modified to contain user details which could cause merge conflict, because user details will only be implemented in the sign up backend issue.

Should I remove bcrypt module and user service create() modifications from my branch for now and let @suprunOksana implement it? Then I could finish password verification in sign in flow when sign up backend will be merged with main.

apps/backend/src/modules/auth/libs/enums/enums.ts
```diff
@@ -1 +1 @@
-export { AuthApiPath } from "@meetlytic/shared";
+export { AuthApiPath, AuthStausMessage } from "@meetlytic/shared";
```
VolodymyrKozel, line 1: Hey! 👋
I noticed a small typo in the export from @meetlytic/shared – it was AuthStausMessage instead of AuthStatusMessage.

apps/backend/src/libs/modules/bcrypt/base-bcrypt.module.ts
```diff
@@ -0,0 +1,19 @@
+import bcrypt from "bcrypt";
+
+class BaseBcrypt {
+	async hash(password: string): Promise<{ hash: string; salt: string }> {
+		const salt = await bcrypt.genSalt();
```
fshabanov, line 5: Let's pass salt rounds here. Add salt rounds to .env

apps/backend/src/modules/auth/libs/exceptions/auth.error.ts
```diff
@@ -0,0 +1,15 @@
+import { HTTPCode } from "~/libs/enums/enums.js";
+import { HTTPError } from "~/libs/exceptions/exceptions.js";
+
+import { AuthStausMessage } from "../enums/enums.js";
+
+class AuthError extends HTTPError {
```
fshabanov, line 6: Should be in shared

apps/backend/src/modules/auth/libs/exceptions/auth.error.ts
```diff
@@ -0,0 +1,15 @@
+import { HTTPCode } from "~/libs/enums/enums.js";
+import { HTTPError } from "~/libs/exceptions/exceptions.js";
+
+import { AuthStausMessage } from "../enums/enums.js";
```
fshabanov, line 4:
```suggestion
import { AuthStatusMessage } from "../enums/enums.js";
```

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,16 +1,41 @@
...
+		}
+		const isValid = await bcrypt.verify(
+			userRequestDto.password,
+			user.passwordHash,
+		);
+		if (!isValid) {
+			throw new AuthError();
+		}
+		return {
+			token: "token",
```
fshabanov, line 34: Do not return anything hardcoded

apps/backend/src/modules/users/user.entity.ts
```diff
@@ -64,11 +64,13 @@ class UserEntity implements Entity {
 
 	public toNewObject(): {
 		email: string;
+		id: number;
```
fshabanov, line 67: `toNewObject` should not receive an id

apps/backend/src/modules/users/user.service.ts
```diff
@@ -1,3 +1,4 @@
+import { bcrypt } from "~/libs/modules/bcrypt/bcrypt.js";
```
fshabanov, line 1: The name `bcrypt` is conflicting with the library name, let's name the module `encrypt`

apps/backend/src/modules/users/user.service.ts
```diff
@@ -33,8 +35,12 @@ class UserService implements Service {
 		return Promise.resolve(true);
 	}
 
-	public find(): ReturnType<Service["find"]> {
-		return Promise.resolve(null);
+	public async find(
+		id: number,
+	): Promise<null | ReturnType<UserEntity["toNewObject"]>> {
```
fshabanov, line 40: `find` should not return `toNewObject` ReturnType. We should have one common type `UserAuthResponseDto` and return it

apps/backend/src/modules/users/user.service.ts
```diff
@@ -45,6 +51,14 @@ class UserService implements Service {
 		};
 	}
 
+	public async findByEmail(
+		email: string,
+	): Promise<null | ReturnType<UserEntity["toNewObject"]>> {
```
fshabanov, line 56: Same here

fshabanov (changes requested):

apps/backend/src/modules/auth/auth.service.ts
Karpt221: But I cannot return the actual token while jwt module is not ready yet. Can it be here until then?

fshabanov: If there's no jwt module yet, do not return token at all. If we merge jwt module before your PR, you will add it later

VolodymyrKozel: The token should be generated and returned by sign-in and sign-up endpoints it is task #9

sofiia-trokhymchuk:
Fix conflicts

apps/backend/src/libs/modules/encrypt/base-encrypt.module.ts
```diff
@@ -0,0 +1,21 @@
...
+
+		return { hash, salt };
+	}
+
+	async verify(
+		plainPassword: string,
+		hashedPassword: string,
+	): Promise<boolean> {
+		return await bcrypt.compare(plainPassword, hashedPassword);
+	}
```
fshabanov, line 18: Shouldn't you also pass salt from DB when comparing?

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -38,6 +41,30 @@ class UserRepository implements Repository {
...
+	public async findByEmail(email: string): Promise<null | UserEntity> {
+		const user = await this.userModel
+			.query()
+			.findOne(UserAttribute.EMAIL, email);
+		return user ? UserEntity.initialize(user) : null;
+	}
+
+	public async getCredentials(id: number): Promise<null | {
+		passwordHash: string;
+		passwordSalt: string;
```
fshabanov, line 53: I noticed you use this type in multiple places. Let's create a separate type

fshabanov (changes requested):

apps/backend/src/libs/modules/encrypt/base-encrypt.module.ts
fshabanov: Let's keep the current approach for now

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -38,6 +42,27 @@ class UserRepository implements Repository {
...
+			.query()
+			.select(UserAttribute.PASSWORD_HASH, UserAttribute.PASSWORD_SALT)
+			.findById(id);
+
+		return credentials
+			? {
+					passwordHash: credentials.passwordHash,
+					passwordSalt: credentials.passwordSalt,
+				}
+			: null;
```
fshabanov, line 63:
```suggestion
		return credentials ?? null;
```

fshabanov (approved):

fshabanov:
Merge the main branch into yours

apps/backend/src/libs/modules/encrypt/base-encrypt.module.ts
```diff
@@ -0,0 +1,21 @@
+import bcrypt from "bcrypt";
+
+class BaseEncrypt {
+	constructor(public saltRounds: number) {}
```
sofiia-trokhymchuk, line 4: let's stick to one code style
```suggestion
  public saltRounds: number;
  
	constructor(saltRounds: number) {
	  this.saltRounds = saltRounds;
	}
```

o-nedashkivska (approved):

apps/backend/src/modules/users/user.service.ts
```diff
@@ -42,6 +47,16 @@ class UserService implements Service {
 			items: items.map((item) => item.toObject()),
 		};
 	}
+	public async findByEmail(email: string): Promise<null | UserResponseDto> {
+		const user = await this.userRepository.findByEmail(email);
+
+		return user ? user.toObject() : null;
+	}
```
sofiia-trokhymchuk, line 54: find by id and by email seem to be almost the same. consider refactoring your code to make the find method as generic as possible. it should be a find method that accepts partial properties of user type

sofiia-trokhymchuk (changes requested):

apps/backend/src/modules/users/user.service.ts
sofiia-trokhymchuk: if find method was agreed to use id only then let's leave it as it is 🙌

sofiia-trokhymchuk (approved):

PR #23, opened 2025-07-21
feat: add protected routing ml-10

ulianchenko:


ulianchenko:
Temporarily, while the user.service and jwt.service are not ready, I’m using stubs. Once the services are ready, I’ll refactor the code.

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,100 @@
...
+	(fastify, options: Options, done) => {
+		const { routesWhiteList, services } = options;
+		const { jwt, logger, userService } = services;
+
+		fastify.decorateRequest("user", null);
+
+		fastify.addHook("preHandler", async (request) => {
+			const { url } = request;
+
+			if (routesWhiteList.includes(url)) {
```
VolodymyrKozel, line 38: In your current implementation,
```ts
if (routesWhiteList.includes(url)) {
			return;
			}
```
you only checks the URL, but not the HTTP method. This is a potential security issue because the same path (e.g., /product) can have different access rules depending on the method:
    GET /product – public (allowed without authentication)
    POST /product – protected (requires authentication)

To handle this correctly, you should create a function that checks both the method and URL against your whitelist.

fshabanov:
> Temporarily, while the user.service and jwt.service are not ready, I’m using stubs. Once the services are ready, I’ll refactor the code.

You can merge jwt implementation to your branch. Regarding user service, you should implement it yourself

ulianchenko:
Regarding user service, @Karpt221 already implemented user service and find(id) method that I need already done. Should I write another implementation or rewrite existing implementation of user service?

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,99 @@
...
+
+			try {
+				const { payload } = await jwt.verify(token);
+
+				const userId = payload.userId;
+
+				const user = await userService.findById(userId);
+
+				if (!user) {
+					throw new HTTPError({
```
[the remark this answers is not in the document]
Karpt221: I have already implemented AuthError.

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,99 @@
...
+					status: HTTPCode.UNAUTHORIZED,
+				});
+			}
+
+			try {
+				const { payload } = await jwt.verify(token);
+
+				const userId = payload.userId;
+
+				const user = await userService.findById(userId);
```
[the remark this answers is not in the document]
Karpt221: In my implementation of userService I kept the default find() method instead of findById() to align with interface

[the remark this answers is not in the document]
ulianchenko: In my last commit I rewrote to use your implementation of the userService with some changes.
```
public async find(
		id: number,
	): Promise<null | ReturnType<UserEntity["toNewObject"]>> {
		const user = await this.userRepository.find(id);

		return user ? user.toNewObject() : null;
	}
```

changed to:

```
public async find(id: number): Promise<null | UserSignUpResponseDto> {
		const user = await this.userRepository.find(id);

		return user ? user.toObject() : null;
	}
```

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -67,10 +68,7 @@ const authorizationPlugin: FastifyPluginCallback<Options> = fp(
 				const user = await userService.find(userId);
 
 				if (!user) {
-					throw new HTTPError({
-						message: "User not found.",
-						status: HTTPCode.UNAUTHORIZED,
-					});
+					throw new AuthError();
```
Karpt221, line 71: You can use `AuthError`  when checking authorizationHeader and token, instead of `HTTPError`

apps/backend/src/libs/constants/white-routes.constant.ts
```diff
@@ -0,0 +1,10 @@
+import { APIPath, AuthApiPath } from "@meetlytic/shared";
```
fshabanov, line 1:
```suggestion
import { APIPath, AuthApiPath } from "~/libs/enums/enums.js";
```

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -58,6 +62,8 @@ class BaseServerApplication implements ServerApplication {
 		this.app = Fastify({
 			ignoreTrailingSlash: true,
 		});
+
+		this.app.decorate(FastifyDecorator.API_VERSIONS, new Set<string>());
```
fshabanov, line 66: Why do we need this?

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -209,8 +213,20 @@ class BaseServerApplication implements ServerApplication {
...
 		);
+
+		await this.app.register(authorizationPlugin, {
+			routesWhiteList: WHITE_ROUTES,
+			services: {
+				config: this.config,
+				jwt: jwt,
+				logger: this.logger,
+				userService: userService,
+			},
```
fshabanov, line 228:
```suggestion
			services: {
				config: this.config,
				jwt,
				logger: this.logger,
				userService,
			},
```

fshabanov (changes requested):

apps/backend/src/libs/modules/server-application/base-server-application.ts
fshabanov: I think there should be a possibility to get only the endpoint from the request (like /auth/sign-in), we can use it instead

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -5,17 +5,21 @@ import Fastify, { type FastifyError, type FastifyInstance } from "fastify";
 import path from "node:path";
 import { fileURLToPath } from "node:url";
 
+import { WHITE_ROUTES } from "~/libs/constants/white-routes.constant.js";
```
fshabanov, line 8:
```suggestion
import { WHITE_ROUTES } from "~/libs/constants/constants.js";
```

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,96 @@
...
+
+import { type WhiteRoute } from "~/libs/constants/libs/types/types.js";
+import { FastifyHook } from "~/libs/enums/enums.js";
+import { AuthError } from "~/libs/exceptions/exceptions.js";
+import { type Config } from "~/libs/modules/config/config.js";
+import { type Logger } from "~/libs/modules/logger/logger.js";
+import { type BaseToken } from "~/libs/modules/token/base-token.module.js";
+import { type JwtPayload } from "~/libs/modules/token/libs/types/types.js";
+import { type UserResponseDto } from "~/modules/users/libs/types/types.js";
+import { type UserService } from "~/modules/users/user.service.js";
```
fshabanov, line 12:
```suggestion
import { type BaseToken, type JwtPayload } from "~/libs/modules/token/token.js";
import { type UserService, type UserResponseDto } from "~/modules/users/user.service.js";
```

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,96 @@
...
+
+declare module "fastify" {
+	interface FastifyRequest {
+		user?: null | UserResponseDto;
+	}
+}
```
fshabanov, line 18: Move to a separate .d.ts file

fshabanov (changes requested):

fshabanov (approved):

PR #35, opened 2025-07-22
feat: add loader component ml-13

MrKopchick:


apps/frontend/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,25 @@
...
+	isLoading: boolean;
+};
+
+const Loader: React.FC<Properties> = ({ isLoading }) => {
+	if (!isLoading) {
+		return null;
+	}
+
+	return (
+		<div className="loader-overlay">
```
o-nedashkivska, line 15: We might also need loader without overlay

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -27,6 +29,8 @@ const App: React.FC = () => {
 
 	return (
 		<>
+			<Loader isLoading={isLoading} />
```
o-nedashkivska, line 32: For now we don't need loader here

MrKopchick: I added this here to trivially fix the lint:trash error. It's more of a temporary option. If it's not used here, then where?

sofiia-trokhymchuk: for demo purposes you can check if `dataStatus` is `Pending` for `userActions.loadAll` for example

apps/frontend/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,30 @@
...
+			<span />
+			<span />
+			<span />
+		</div>
+	);
+
+	return withOverlay ? (
+		<div className="loader-overlay">{loaderContent}</div>
+	) : (
+		loaderContent
```
sofiia-trokhymchuk, line 26: consider using `clsx` as we agreed to add it to the project

sofiia-trokhymchuk (changes requested):

apps/frontend/src/libs/components/loader/loader.css
```diff
@@ -0,0 +1,48 @@
...
+	background-color: rgb(255 255 255 / 60%);
+	backdrop-filter: blur(0);
+}
+
+.dot-loader {
+	display: flex;
+	gap: 8px;
+}
+
+.dot-loader span {
```
fshabanov, line 20: B10

apps/frontend/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,30 @@
...
+	const loaderContent = (
+		<div className="dot-loader">
+			<span />
+			<span />
+			<span />
+		</div>
+	);
+
+	return withOverlay ? (
+		<div className="loader-overlay">{loaderContent}</div>
```
fshabanov, line 24: Use css modules instead

fshabanov, line 24: Use css modules instead

apps/frontend/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,27 @@
+import { clsx } from "clsx";
```
fshabanov, line 1: Create a common helper `getValidClassNames`

fshabanov (changes requested):

fshabanov (approved):

o-nedashkivska (approved):
When you request a review, don't forget to change the status of the ticket to `In Review`

PR #36, opened 2025-07-22
feat: add sidebar component ml-17

VolodymyrKozel:
example of usage:

```ts
import { NAVIGATION_ITEMS } from "~/libs/constants/constants.js";

     <Sidebar>
         <Navigation items={NAVIGATION_ITEMS} />
     </Sidebar>

### CODE REVIEW

**Review by o-nedashkivska — COMMENTED** [review]
All fonts and variables from the design should be added in a separate task. Make sure to only add styles that you need for current task

**Comment by VolodymyrKozel** [comment]
> All fonts and variables from the design should be added in a separate task. Make sure to only add styles that you need for current task
hi, i [fix: remove unused variable and fix css linting issues ml-17](https://github.com/BinaryStudioAcademy/bsa-2025-meetlytic/pull/36/commits/9ed6c1a683d5315dbfc416ad79185675e494881d)

File: apps/frontend/src/libs/components/navigation/libs/components/navigation-link/navigation-link.tsx [hunk 1]
```diff
@@ -0,0 +1,38 @@
...
+}: Properties) => {
+	const getClassName = useCallback(
+		({ isActive }: { isActive: boolean }): string => {
+			return getValidClassNames(
+				styles["navigation-link"],
+				isActive && styles["active"],
+			);
+		},
+		[],
+	);
```
fshabanov, line 21: I don't think we need this additional function, you can pass getValidClassNames() directly to component

apps/frontend/src/assets/img/icons/integrations.svg
fshabanov, line 1: I don't think we use this icon for now. Only add the ones you need

fshabanov (changes requested):

apps/frontend/src/libs/components/navigation/libs/components/navigation-link/navigation-link.tsx
VolodymyrKozel: Hi @fshabanov ,
Thanks for the suggestion! 🙌
I tried passing the getValidClassNames function directly as an inline arrow function like this:
```ts
<NavLink
	className={({ isActive }) =>
		getValidClassNames(
			styles["navigation-link"],
			isActive && styles["active"]
		)
	}
	to={href}
	key={label}
/>
```
However, this triggers the ESLint react/jsx-no-bind rule, which warns against using arrow functions directly in JSX props for performance reasons.
To follow the linter rules and keep clean code, I moved the function outside the JSX:
```ts
const getClassName = ({ isActive }: { isActive: boolean }): string =>
	getValidClassNames(styles["navigation-link"], isActive && styles["active"]);
```
Then passed it like this:
```ts
<NavLink className={getClassName} to={href} key={label} />
```
Let me know if you'd prefer another approach!
Thanks again for the feedback 🙏

fshabanov (approved):

PR #37, opened 2025-07-22
feat: add protected routing ml-11

Roman-PRV:
- [x] Create ProtectedRoute wrapper component
- [x] Redirect not signed in users from protected screens to Sign In page

Please note: using this branch activates protected routing, which means you will not be able to access any pages except for login and registration until the relevant authorization issues are activated.

### DIFF
--- a/apps/frontend/src/index.tsx
+++ b/apps/frontend/src/index.tsx
@@ -11,27 +11,29 @@ import { AppRoute } from "~/libs/enums/enums.js";
 import { store } from "~/libs/modules/store/store.js";
 import { Auth } from "~/pages/auth/auth.jsx";
 
+import { ProtectedRoute } from "./libs/components/router-provider/protected-route.js";
+
 createRoot(document.querySelector("#root") as HTMLElement).render(
 	<StrictMode>
 		<StoreProvider store={store.instance}>
 			<RouterProvider
 				routes={[
 					{
+						element: <Auth />,
+						path: AppRoute.SIGN_IN,
+					},
+					{
+						element: <Auth />,
+						path: AppRoute.SIGN_UP,
+					},
+					{
 						children: [
 							{
-								element: "Root",
+								element: <App />,
 								path: AppRoute.ROOT,
 							},
-							{
-								element: <Auth />,
-								path: AppRoute.SIGN_IN,
-							},
-							{
-								element: <Auth />,
-								path: AppRoute.SIGN_UP,
-							},
 						],
-						element: <App />,
+						element: <ProtectedRoute />,
 						path: AppRoute.ROOT,
 					},
 				]}
--- a/apps/frontend/src/libs/components/router-provider/protected-route.tsx
+++ b/apps/frontend/src/libs/components/router-provider/protected-route.tsx
@@ -1,0 +1,24 @@
+import { FC } from "react";
+import { Navigate, Outlet, useLocation } from "react-router-dom";
+
+import { AppRoute } from "~/libs/enums/enums.js";
+import { useAppSelector } from "~/libs/hooks/hooks.js";
+
+const ProtectedRoute: FC = () => {
+	const isAuthenticated = useAppSelector((state) => state.auth.isAuthenticated);
+	const location = useLocation();
+
+	if (!isAuthenticated) {
+		return (
+			<Navigate
+				replace
+				state={{ from: location.pathname }}
+				to={AppRoute.SIGN_IN}
+			/>
+		);
+	}
+
+	return <Outlet />;
+};
+
+export { ProtectedRoute };
--- a/apps/frontend/src/libs/components/router-provider/protected-route.tsx
+++ b/apps/frontend/src/libs/components/router-provider/protected-route.tsx
@@ -5,7 +5,7 @@ import { AppRoute } from "~/libs/enums/enums.js";
 import { useAppSelector } from "~/libs/hooks/hooks.js";
 
 const ProtectedRoute: FC = () => {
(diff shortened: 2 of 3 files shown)


apps/frontend/src/libs/components/router-provider/protected-route.tsx
```diff
@@ -0,0 +1,24 @@
+import { FC } from "react";
+import { Navigate, Outlet, useLocation } from "react-router-dom";
+
+import { AppRoute } from "~/libs/enums/enums.js";
+import { useAppSelector } from "~/libs/hooks/hooks.js";
+
+const ProtectedRoute: FC = () => {
```
fshabanov, line 7: Should receive children that need to be displayed. Also, add a prop `redirectTo`, which will be AppRoute

apps/frontend/src/modules/auth/slices/auth.slice.ts
```diff
@@ -7,10 +8,14 @@ import { signUp } from "./actions.js";
 
 type State = {
 	dataStatus: ValueOf<typeof DataStatus>;
+	token: null | string;
```
fshabanov, line 11: You don't need to save token in state. We already have it in localStorage

apps/frontend/src/modules/auth/slices/auth.slice.ts
```diff
@@ -1,3 +1,4 @@
+import { UserResponseDto } from "@meetlytic/shared";
```
fshabanov, line 1:
```suggestion
import { type UserResponseDto } from "~/libs/types/types.js";
```

fshabanov (changes requested):

apps/frontend/src/libs/components/router-provider/protected-route.tsx
```diff
@@ -0,0 +1,29 @@
+import { ValueOf } from "@meetlytic/shared";
+import { FC, ReactNode } from "react";
+import { Navigate, useLocation } from "react-router-dom";
+
+import { AppRoute } from "~/libs/enums/enums.js";
+import { useAppSelector } from "~/libs/hooks/hooks.js";
+
+type Propertis = {
```
o-nedashkivska, line 8:
```suggestion
type Properties = {
```

apps/frontend/src/index.tsx
```diff
@@ -12,28 +12,34 @@ import { AppRoute } from "~/libs/enums/enums.js";
 import { store } from "~/libs/modules/store/store.js";
 import { Auth } from "~/pages/auth/auth.jsx";
 
+import { ProtectedRoute } from "./libs/components/router-provider/protected-route.js";
```
fshabanov, line 15:
```suggestion
import { ProtectedRoute } from "./libs/components/components.js";
```

apps/frontend/src/libs/components/router-provider/protected-route.tsx
```diff
@@ -0,0 +1,29 @@
+import { ValueOf } from "@meetlytic/shared";
```
fshabanov, line 1:
```suggestion
import { ValueOf } from "~/libs/types/types.js";
```

apps/frontend/src/libs/components/router-provider/protected-route.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { FC, ReactNode } from "react";
+import { Navigate, useLocation } from "react-router-dom";
```
fshabanov, line 3:
```suggestion
import { Navigate } from "~/libs/components/components.js"
import { useLocation } from "~/libs/hooks/hooks.js"
```

fshabanov (changes requested):

apps/frontend/src/index.tsx
```diff
@@ -12,28 +12,34 @@ import { AppRoute } from "~/libs/enums/enums.js";
...
-								path: AppRoute.SIGN_IN,
-							},
-							{
-								element: <Auth />,
-								path: AppRoute.SIGN_UP,
-							},
 						],
-						element: <App />,
+						element: (
+							<ProtectedRoute redirectTo={AppRoute.SIGN_IN}>
```
o-nedashkivska, line 39: You can omit `redirectTo` prop here as it is defaults to `AppRoute.SIGN_IN`
```suggestion
							<ProtectedRoute>
```

fshabanov: I prefer we add redirectTo explicitly, in case we want to redirect to other screens at some time, not only the default one

Roman-PRV: It probably makes sense to remove the default value if we always specify it explicitly.

apps/frontend/src/libs/components/router-provider/protected-route.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { AppRoute } from "~/libs/enums/enums.js";
+import { useAppSelector, useLocation } from "~/libs/hooks/hooks.js";
+import { ValueOf } from "~/libs/types/types.js";
+
+type Properties = {
+	children: ReactNode;
+	redirectTo: ValueOf<typeof AppRoute>;
+};
+
+const ProtectedRoute: FC<Properties> = ({ children, redirectTo }) => {
```
fshabanov, line 13:
```suggestion
type Properties = {
	children: React.ReactNode;
	redirectTo: ValueOf<typeof AppRoute>;
};

const ProtectedRoute: React.FC<Properties> = ({ children, redirectTo }) => {
```
Let's use the same approach everywhere

fshabanov (approved):

PR #38, opened 2025-07-23
feat: add shared styles ml-33

yehor-kostiuk:
* Typography
* Colors
* Inputs
* Reset styles


apps/frontend/src/assets/css/variables.css
```diff
@@ -0,0 +1,182 @@
+.h1,
```
fshabanov, line 1: Those are not variables tho
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascading_variables/Using_CSS_custom_properties

apps/frontend/src/assets/css/variables.css
```diff
@@ -0,0 +1,182 @@
...
+	font-weight: 700;
+	line-height: 100%;
+}
+
+.field {
+	width: 100%;
+	margin-block: 0;
+}
+
+.field label {
```
fshabanov, line 88: B10

apps/frontend/src/assets/css/variables.css
```diff
@@ -0,0 +1,182 @@
...
+	--color-background-3: rgb(246 245 245);
+	--color-background-4: rgb(255 255 255);
+	--color-text-1: rgb(41 45 50);
+	--color-text-2: rgb(175 175 175);
+	--color-brand: rgb(16 173 216);
+	--color-brand-hover: rgb(15 143 178);
+	--color-accent-1: rgb(255 94 94);
+	--color-accent-2: rgb(170 235 152);
+	--border-radius: 5px;
+}
```
fshabanov, line 182: These are variables. Everything else in this file is not

VolodymyrKozel: hi, where is fonts?

apps/frontend/src/assets/css/scaffolding.css
```diff
@@ -1,14 +1,75 @@
...
+}
+
 body {
 	display: flex;
 	flex-direction: column;
 	min-width: 320px;
 	min-height: 100vh;
 	margin: 0;
-	font-family: Arial, sans-serif;
+	font-family: Inter, sans-serif;
```
VolodymyrKozel, line 72: why we do not use variables for fonts.

apps/frontend/src/assets/css/typography.css
```diff
@@ -0,0 +1,77 @@
...
+	font-weight: 700;
+	line-height: 28.06px;
+	letter-spacing: 0;
+}
+
+.text-h5 {
+	font-size: 16px;
+	font-weight: 600;
+	line-height: 25px;
+	letter-spacing: 0;
```
VolodymyrKozel, line 33: do we need override the inherited spacing.

apps/frontend/src/assets/css/variables.css
yehor-kostiuk: i see Inter and Nunito in figma , but cant find others

yehor-kostiuk: In most cases, Inter and Nunito are used, other fonts are designations or inscriptions in the design, but I will add them if necessary.

apps/frontend/src/assets/css/typography.css
```diff
@@ -0,0 +1,66 @@
+.text-h1 {
```
fshabanov, line 1: For typography you only need to setup fonts, no need to create these classes

apps/frontend/src/assets/css/form.css
```diff
@@ -0,0 +1,47 @@
+.form-field {
```
fshabanov, line 1: Component specific styles will be added separately for components, not as global styles

apps/frontend/src/assets/css/button.css
```diff
@@ -0,0 +1,38 @@
+.btn-primary {
```
fshabanov, line 1: Same here

apps/frontend/src/assets/css/typography.css
yehor-kostiuk: or with styling , but without classes just a h1 h2 h3

packages/shared/src/modules/users/libs/enums/user-validation-message.enum.ts
```diff
@@ -1,4 +1,4 @@
-import { UserValidationRule } from "../enums/enums.js";
+import { UserValidationRule } from "../enums/user-validation-rule.enum.js";
```
fshabanov, line 1:
```suggestion
import { UserValidationRule } from "./user-validation-rule.enum.js";
```

fshabanov (approved):

PR #47, opened 2025-07-24
feat: add reusable button component ml-39

yehor-kostiuk:


apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -1,11 +1,37 @@
+import { getValidClassNames } from "~/libs/helpers/helpers.js";
+
+import styles from "./styles.module.css";
+
 type Properties = {
+	disabled?: boolean;
```
fshabanov, line 6: B19

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -1,11 +1,37 @@
...
+	onClick,
+	size = "default",
 	type = "button",
-}: Properties) => <button type={type}>{label}</button>;
+	variant = "primary",
+}: Properties) => (
+	<button
+		className={getValidClassNames(
+			styles["button"],
+			styles[`button--${variant}`],
```
fshabanov, line 25:
```suggestion
			styles[`button-${variant}`],
```

o-nedashkivska:
> > > what about border-radius, in figma its 2 main values 5px 10px and in few places 8px, better to create variables for it ? or everyone will do it in components. Since I implemented the radius variable only for 5px, and in some places we use var, in others only for px, I think we should discuss a common structure, it will also be easier to change the radius variable if we want to change it everywhere.
> > 
> > 
> > We can have border radius sizes like --border-radius-sm: 5px; --border-radius-md: 8px; ...
> 
> whats about 10px ? it uses in summary cards in in some pop-up windows and in some places more

You can add sizes for any border-radius that is reused throughout the app

sofiia-trokhymchuk:
Consider adding children to Props or appended/prepended components (right/left) to pass not only labels but also icons

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -1,11 +1,37 @@
+import { getValidClassNames } from "~/libs/helpers/helpers.js";
+
+import styles from "./styles.module.css";
+
 type Properties = {
+	isDisabled?: boolean;
 	label: string;
+	onClick?: React.MouseEventHandler<HTMLButtonElement>;
```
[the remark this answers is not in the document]
yehor-kostiuk: what to do if it is submit type, create a stub? I just thought it was not a very good option. In the form we don't need onClick because the button just submits the form

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -1,11 +1,37 @@
...
+	size?: "default" | "small";
 	type?: "button" | "submit";
+	variant?: "outlined" | "primary";
```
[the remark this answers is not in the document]
yehor-kostiuk: I did it, but I think it became a little harder to understand with more files and more code and why we don't use enum for types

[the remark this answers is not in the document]
yehor-kostiuk: check now, if it's good - leave it

fshabanov (changes requested):

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -1,11 +1,49 @@
...
+	variant?: ButtonVariant;
 };
 
 const Button: React.FC<Properties> = ({
+	iconLeft,
+	iconRight,
+	isDisabled = false,
 	label,
+	onClick,
+	size = BUTTON_SIZE.DEFAULT,
```
fshabanov, line 28: B23

apps/frontend/src/libs/enums/button-size.enum.ts
```diff
@@ -0,0 +1,9 @@
+const BUTTON_SIZE = {
+	DEFAULT: "default",
+	SMALL: "small",
+} as const;
+
+type ButtonSize = (typeof BUTTON_SIZE)[keyof typeof BUTTON_SIZE];
```
fshabanov, line 6:
```suggestion
type ButtonSize = ValueOf<typeof BUTTON_SIZE>
```
Also, do not export a type from enums folder

fshabanov (changes requested):

yehor-kostiuk:
SCR-20250725-rmvw

PR #49, opened 2025-07-24
feat: add aws integration ml-31

VolodymyrKozel:
This PR introduces support for creating and deleting AWS EC2 instances using CloudFormation via the @aws-sdk/client-cloudformation.

**Key parts of the implementation include:**
    - Environment variables for AMI_ID and SECURITY_GROUP_ID
    - Updates to the config module to access these AWS-related variables
    - CloudFormation JSON template created for EC2 instance deployment
    - A dedicated module (CloudFormationEC2) to manage stack creation and deletion
    - Initialization logic in a single entry point file


apps/backend/src/libs/modules/ec2-cloudformation/ec2-cloudformation.ts
```diff
@@ -0,0 +1,20 @@
+import { config } from "~/libs/modules/config/config.js";
+import { logger } from "~/libs/modules/logger/logger.js";
+
+import { CloudFormationEC2 } from "./ec2-cloudformation.module.js";
+import templateJSON from "./ec2-instance-template.json" with { type: "json" };
```
VolodymyrKozel, line 5: i have error "Unexpected use of file extension "json" for "./ec2-instance-template.json" eslint. 
Solutions:
- create rule and allow json extention.
- use fs to read file.

apps/backend/src/libs/modules/ec2-cloudformation/ec2-cloudformation.ts
```diff
@@ -0,0 +1,20 @@
...
+import { AWSRegion } from "./libs/enums/enums.js";
+
+const {
+	AWS: { ACCESS_KEY_ID, SECRET_ACCESS_KEY },
+} = config.ENV;
+
+const ec2 = new CloudFormationEC2({
+	accessKeyId: ACCESS_KEY_ID,
+	logger,
+	region: AWSRegion.EU_NORTH_1,
```
fshabanov, line 15: Region should also be set from the .env

apps/backend/src/libs/modules/ec2-cloudformation/ec2-cloudformation.module.ts
```diff
@@ -0,0 +1,129 @@
...
+		this.#client = new CloudFormationClient({
+			credentials: {
+				accessKeyId: accessKeyId,
+				secretAccessKey: secretAccessKey,
+			},
+			region,
+		});
+	}
+
+	async #getInstanceIdFromStack(stackName: string): Promise<string> {
```
fshabanov, line 56:
```suggestion
	private async getInstanceIdFromStack(stackName: string): Promise<string> {
```

Use the same approach for all

VolodymyrKozel: hey, ts run on dev and compile to the js so private functions and variables convert to public.  I change # to private same to variables.

fshabanov: I think we don't need to worry about the case you mentioned. Even if ts compiles private methods to public, it will not allow us to compile if we try to use private methods outside the class

apps/backend/.env.example
```diff
@@ -24,3 +24,12 @@ DB_POOL_MAX=10
...
 SALT_ROUNDS=10
+
+#
+# AWS
+#
+ACCESS_KEY_ID=
+AMI_ID=
+REGION=
+SECRET_ACCESS_KEY=
+SECURITY_GROUP_ID=
```
fshabanov, line 35: As we discussed, let's create security group using cloudformation, instead of passing the id

apps/backend/src/libs/modules/ec2-cloudformation/ec2-cloudformation.module.ts
```diff
@@ -0,0 +1,125 @@
+import {
```
fshabanov, line 1: Let's rename this module to `cloud-formation`

apps/backend/src/libs/modules/ec2-cloudformation/ec2-cloudformation.module.ts
```diff
@@ -0,0 +1,125 @@
...
+	CloudFormationClient,
+	CreateStackCommand,
+	DeleteStackCommand,
+	DescribeStacksCommand,
+	waitUntilStackCreateComplete,
+	waitUntilStackDeleteComplete,
+} from "@aws-sdk/client-cloudformation";
+
+import { meetingService } from "~/modules/meetings/meetings.js";
```
fshabanov, line 10: Pass the service through the constructor as in other places

apps/backend/src/libs/modules/ec2-cloudformation/ec2-cloudformation.module.ts
```diff
@@ -0,0 +1,125 @@
...
+	private async getInstanceIdFromStack(stackName: string): Promise<string> {
+		const describeCommand = new DescribeStacksCommand({ StackName: stackName });
+		const { Stacks } = await this.client.send(describeCommand);
+		const stack = Stacks?.find((stack) => stack.StackName === stackName);
+		const instanceId = stack?.Outputs?.find(
+			(output) => output.OutputKey === OutputKey.INSTANCE_ID,
+		)?.OutputValue;
+
+		if (!instanceId) {
+			throw new Error("Failed to get InstanceId from CloudFormation output");
```
fshabanov, line 66: Error messages should be in enum ExceptionMessage

apps/backend/src/modules/meetings/meeting.service.ts
```diff
@@ -1,3 +1,4 @@
+import { ec2 } from "~/libs/modules/ec2-cloudformation/ec2-cloudformation.js";
```
fshabanov, line 1: Through constructor

eslint.config.js
```diff
@@ -76,8 +76,12 @@ const importConfig = {
 		"import/exports-last": ["error"],
 		"import/extensions": [
 			"error",
+			"ignorePackages",
 			{
-				js: "ignorePackages",
+				js: "always",
+				json: "always",
+				ts: "never",
+				tsx: "never",
```
fshabanov, line 84:
```suggestion
				js: "ignorePackages",
				json: "always",
```
Would that work?

packages/shared/src/modules/meetings/libs/enums/meeting-status-message.enum.ts
```diff
@@ -1,6 +1,7 @@
 const MeetingStatusMessage = {
 	CANNOT_UPDATE_NON_EXISTENT: "Cannot update non-existent meeting with ID",
 	FORBIDDEN: "Access denied",
+	MEETING_FAILD_TO_CREATE: "Failed to create meeting",
```
fshabanov, line 4:
```suggestion
	MEETING_FAILED_TO_CREATE: "Failed to create a meeting",
```

fshabanov (changes requested):

eslint.config.js
VolodymyrKozel: work

apps/backend/src/libs/modules/cloud-formation/cloud-formation.module.ts
```diff
@@ -0,0 +1,121 @@
...
+type Constructor = {
+	credentials: {
+		accessKeyId: string;
+		secretAccessKey: string;
+	};
+	imageId: string;
+	logger: Logger;
+	meetingService: MeetingService;
+	region: string;
+	templateBody: string;
```
fshabanov, line 30: Let's pass templateBody to the create method directly to have a possibility to use different templates if required

apps/backend/src/libs/modules/cloud-formation/libs/enums/cloudformation-exception-message.enum.ts
```diff
@@ -0,0 +1,5 @@
+const ExceptionMessage = {
```
fshabanov, line 1: I think we already have ExceptionMessage enum. Use it

apps/backend/src/libs/modules/cloud-formation/libs/enums/enums.ts
```diff
@@ -0,0 +1,5 @@
+export { Capability } from "./cloudformation-capability.enum.js";
```
fshabanov, line 1:
```suggestion
export { Capability } from "./capability.enum.js";
```
Same for all. Since all files are inside cloud-formation folder, we don't need to specify it again

fshabanov (changes requested):

apps/backend/src/libs/modules/cloud-formation/libs/enums/cloudformation-exception-message.enum.ts
VolodymyrKozel: we have: 
```ts
const MeetingStatusMessage = {
	CANNOT_UPDATE_NON_EXISTENT: "Cannot update non-existent meeting with ID",
	FORBIDDEN: "Access denied",
	MEETING_FAILED_TO_CREATE: "Failed to create a meeting",
	MEETING_NOT_FOUND: "Meeting not found",
	UPDATE_FAILED: "Failed to update meeting",
} as const;

export { MeetingStatusMessage };
```

o-nedashkivska:
If you want to update your branch to be up to date with main but your base branch is not main, first of all, the base branch should be updated with main, then your branch should be updated with base branch.

apps/backend/src/libs/exceptions/cloud-formation.exception.ts
```diff
@@ -0,0 +1,18 @@
+import { ExceptionMessage } from "@meetlytic/shared";
```
fshabanov, line 1:
```suggestion
import { ExceptionMessage } from "~/libs/enums/enums.js";
```
```suggestion
import { ExceptionMessage } from "@meetlytic/shared";
```

apps/backend/src/libs/modules/cloud-formation/cloud-formation.module.ts
```diff
@@ -0,0 +1,140 @@
+import {
+	CloudFormationClient,
+	CreateStackCommand,
+	DeleteStackCommand,
+	DescribeStacksCommand,
+	waitUntilStackCreateComplete,
+	waitUntilStackDeleteComplete,
+} from "@aws-sdk/client-cloudformation";
+
+import { type MeetingService } from "~/modules/meetings/meeting.service.js";
```
fshabanov, line 10:
```suggestion
import { type MeetingService } from "~/modules/meetings/meetings.js";
```

fshabanov (changes requested):

fshabanov (changes requested):

apps/backend/src/modules/meetings/meeting.service.ts
```diff
@@ -82,7 +113,7 @@ class MeetingService implements Service<MeetingResponseDto> {
 		}
 
 		const updatedEntity = MeetingEntity.initialize({
-			host: (payload as MeetingUpdateRequestDto).host,
+			host: (payload as MeetingCreateRequestDto).host,
```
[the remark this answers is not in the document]
VolodymyrKozel: 1. we have type MeetingUpdateRequestDto 
```ts
type MeetingUpdateRequestDto = {
	host: ValueOf<typeof MeetingHost>;
	ownerId: number;
};
```
 you write to @MrKopchick  that we can not update instanceId  from update. Do i need create separate method to udate InstanceId?
2. can we in init as default value pass MeetingHost.ZOOM, instead of ?? MeetingHost.ZOOM,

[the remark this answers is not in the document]
fshabanov: 1. It's not about using a new type, you need it. It is about casting. Partial<T> cannot be casted to T. If you are sure we will always have host there, you can cast host only like `payload.host as ValueOf<typeof MeetingHost>`
2. This is also a possibility

PR #50, opened 2025-07-24
feat: create reusable input component ml-40

Karpt221:
Screenshot from 2025-07-28 20-53-23


apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -29,11 +44,33 @@ const Input = <T extends FieldValues>({
...
+				{iconName && (
+					<Icon
+						className={getValidClassNames(
+							styles["input-icon"],
+							iconClassName,
+							iconPosition && iconPositionToClass[iconPosition],
+						)}
+						name={iconName}
+					/>
+				)}
```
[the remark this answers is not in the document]
Karpt221: Sorry, could you please clarify what benefiets we can get by moving the check logic if the icon name exists to the Icon component?

I wrote this statement with idea to check if user defined iconName at all. Existance of proper icon name accomplished by IconName type. Icon component itself doesn't allow to pass wrong  values or omit the name prop. 

```typescript
type Properties = {
	className?: string | undefined;
	color?: string | undefined;
	name: IconName; 
	// now icon accepts only proper names so there no need to check if it is correct.
	//  It would be strange to allow to skip the name prop
};

const Icon: React.FC<Properties> = ({ className, color, name }: Properties) => {
	const Icon = iconNameToSvg[name];

	return <Icon className={className} color={color} />;
};

```

apps/frontend/src/libs/components/icon/libs/maps/icon-name-to-svg.map.ts
```diff
@@ -1,15 +1,11 @@
 import ArrowRight from "~/assets/img/icons/arrow-right.svg?react";
 import Meetings from "~/assets/img/icons/meetings.svg?react";
 import Settings from "~/assets/img/icons/settings.svg?react";
-import { type IconName } from "~/libs/types/types.js";
 
-const iconNameToSvg: Record<
-	IconName,
-	React.FC<React.SVGProps<SVGSVGElement>>
-> = {
+const iconNameToSvg = {
```
fshabanov, line 5: Using Record<IconName...> would mandate us adding all icons, which is what we need

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -31,9 +46,30 @@ const Input = <T extends FieldValues>({
 	const hasError = Boolean(error);
 
 	return (
-		<label>
-			<span>{label}</span>
-			<input {...field} placeholder={placeholder} type={type} />
+		<label className={styles["lable-wrapper"]}>
```
fshabanov, line 49:
```suggestion
		<label className={styles["label-wrapper"]}>
```

apps/frontend/src/libs/types/icon-name.type.ts
```diff
@@ -1,3 +1,5 @@
-type IconName = "arrowRight" | "meetings" | "settings";
+import { iconNameToSvg } from "~/libs/components/icon/libs/maps/maps.js";
+
+type IconName = keyof typeof iconNameToSvg;
```
fshabanov, line 3: Prefer previous approach

fshabanov (changes requested):

apps/frontend/src/libs/components/input/styles.module.css
```diff
@@ -0,0 +1,46 @@
...
+	border-radius: 5px;
+}
+
+.input:enabled:user-valid {
+	border-color: var(--color-accent-2);
+}
+
+.input:enabled:user-invalid {
+	border-color: var(--color-accent-1);
+}
```
yehor-kostiuk, line 38: Since :user-valid doesn't see React Hook Form errors, instead use hasError and add classes manually, you have full control over validation and can handle all options, because now, even with hasError=true you have a green border, when you delete the entered text the border remains green

apps/frontend/src/libs/components/input/styles.module.css
```diff
@@ -0,0 +1,46 @@
...
+
+.input-icon--right {
+	right: 8px;
+}
+
+.input {
+	width: 100%;
+	padding: 13px 10px;
+	border: 1px var(--color-background-2) solid;
+	border-radius: 5px;
```
yehor-kostiuk, line 29: use variables for radius

yehor-kostiuk:
And there is no "forgot password" option

Karpt221:
Also there is no abbility to handle icon clicks yet.

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -6,36 +6,94 @@ import {
...
+							: isSubmitted && styles["input-valid"],
+						className,
+					)}
+					{...field}
+					id={name}
+					placeholder={placeholder}
+					type={type}
+				/>
+			</div>
 			{hasError && <span>{error as string}</span>}
```
yehor-kostiuk, line 95:
```suggestion
			{hasError && <span className={styles["input-error-label"]}>{error as string}</span>}
			</div>
```

apps/frontend/src/libs/components/avatar/styles.module.css
```diff
@@ -3,21 +3,21 @@
...
-.avatar--main {
+.avatar-main {
 	padding: 3px;
 	border-color: var(--color-brand);
 }
 
-.avatar--alt {
+.avatar-alt {
 	border-color: white;
 }
```
yehor-kostiuk, line 23:
```suggestion
.input-error-label {
	font-size: 10px;
  position: absolute;
  top: -6px;
  z-index: 1;
  left: 2em;
  background-color: var(--color-background-4);
	color: var(--color-accent-1);
  padding: 0 5px;
}
```

yehor-kostiuk, line 23:
```suggestion
.input-error-label {
	font-size: 10px;
  position: absolute;
  top: -6px;
  z-index: 1;
  left: 2em;
  background-color: var(--color-background-4);
	color: var(--color-accent-1);
  padding: 0 5px;
}
```

yehor-kostiuk: It will look something like this 
SCR-20250727-bach

apps/frontend/src/libs/components/input/input.tsx
yehor-kostiuk: and add styles
```
.input-error-label {
  font-size: 10px;
  position: absolute;
  top: -6px;
  z-index: 1;
  left: 32px;
  background-color: var(--color-background-4);
  color: var(--color-accent-1);
  padding: 0 5px;
}
```
It will look something like this
SCR-20250727-bach

Karpt221: I noticed that absolute positioning makes error label overlap label text. I think it will be enough to set only color and font size for now, since design doesn't specify how this element should look like. 

Screenshot from 2025-07-27 09-52-17

```
.input-error-label {
  font-size: 10px;
  background-color: var(--color-background-4);
  color: var(--color-accent-1);
}
```
Screenshot from 2025-07-27 10-03-11

fshabanov: Displaying the error below is enough for now

fshabanov (changes requested):

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -6,36 +6,96 @@ import {
...
-		</label>
+		<div className={wrapperClassName}>
+			<div className={styles["label-wrapper"]}>
+				<label
+					className={hasVisuallyHiddenLabel ? "visually-hidden" : ""}
+					htmlFor={name}
+				>
+					{label}
+				</label>
+				{labelSuffix}
```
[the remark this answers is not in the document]
Karpt221: In the design there are cases where some additional content is needed alongside the label text.
I assumed that this content should always be right of label text. This content can have any structure, so
the separate label guarantees that it will contain only text inside which prevents html errors.
I didn't want to shift the responsibility of HTML validity to the component consumer.

```
<Input
   control={control}
   errors={errors}
   label="Password"
   name="password"
   labelSuffix={<Link to="">Forgot Password?</Link>}
   placeholder="Enter your password"
   type="text"
/>
```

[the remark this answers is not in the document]
fshabanov: Let's remove labelSuffix. If we need Forgot Password, we will add this link separately, it is not a part of input component

apps/frontend/src/libs/components/icon/icon.tsx
```diff
@@ -6,12 +6,25 @@ type Properties = {
 	className?: string | undefined;
 	color?: string | undefined;
 	name: IconName;
+	onClick?: (() => void) | undefined;
```
sofiia-trokhymchuk, line 9: is `undefined` needed as we already specify as `onClick?`?

apps/frontend/src/libs/components/icon/icon.tsx
```diff
@@ -6,12 +6,25 @@ type Properties = {
...
+}: Properties) => {
 	const Icon = iconNameToSvg[name];
+	const isClickable = Boolean(onClick);
 
-	return <Icon className={className} color={color} />;
+	return isClickable ? (
+		<button type="button">
+			<Icon className={className} color={color} onClick={onClick} />
+		</button>
+	) : (
```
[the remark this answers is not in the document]
fshabanov: Yes, remove click logic from the icon component. <button> without creating a new component?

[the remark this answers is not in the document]
fshabanov: But you will be just wrapping it with a simple <button> with not additional styles, why do you need this ConditionalWrap? It looks like too much of an overkill. Just use <button> tag and that should be enough

fshabanov (approved):

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -1,41 +1,112 @@
...
 };
 
 const Input = <T extends FieldValues>({
+	className = "",
 	control,
 	errors,
+	hasVisuallyHiddenLabel = false,
+	iconClassName = "",
+	iconName,
+	iconPosition = iconName && "left",
```
Karpt221, line 38: This condition is redundant

PR #51, opened 2025-07-25
feat: add search input ml-41

ulianchenko:
search

The SearchInput component was also added to the home page to avoid linter errors and should be removed once it's no longer needed.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,86 @@
+name: 🪲 Bug Report
+description: Report a bug
+title: "fix:"
+labels: ["bug"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Please fill out the sections below to help everyone identify and fix the bug
+  - type: textarea
+    id: description
+    attributes:
+      label: Describe your issue
+      placeholder: When I click here this happens
+    validations:
+      required: true
+  - type: textarea
+    id: steps
+    attributes:
+      label: Steps to reproduce
+      placeholder: |
+        1. Go to page X
+        2. Click here
+        3. Look there
+    validations:
+      required: true
+  - type: textarea
+    id: actual
+    attributes:
+      label: What was the actual result?
+      placeholder: This happened
+    validations:
+      required: true
+  - type: textarea
+    id: expected
+    attributes:
+      label: What was the expected result?
+      placeholder: I expected this to happen
+    validations:
+      required: true
+  - type: textarea
+    id: screenshots
+    attributes:
+      label: Put here any screenshots or videos (optional)
+  - type: dropdown
+    id: OS
+    attributes:
+      label: "What OS are you seeing the problem on?"
+      multiple: true
+      options:
+        - "Windows"
+        - "MacOS"
+        - "Android"
+        - "Other (mention in the description)"
+    validations:
+      required: true
+  - type: dropdown
+    id: browser
+    attributes:
+      label: "What browsers are you seeing the problem on?"
+      multiple: true
+      options:
+        - "Chrome"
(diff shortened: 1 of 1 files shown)


apps/frontend/src/libs/hooks/use-app-search/use-app-search.hook.ts
```diff
@@ -0,0 +1,34 @@
...
+	errors: ReturnType<typeof useAppForm<SearchFormValues>>["errors"];
+	handleSearch: () => void;
+	name: keyof SearchFormValues;
+};
+
+const useAppSearch = (
+	callback: (value: string) => void,
+	fieldName: string = "searchField",
+): UseSearchResult => {
+	const { control, errors, getValues } = useAppForm<SearchFormValues>({
```
fshabanov, line 20: This hook shouldn't return control, errors and other things. It should receive the callback, parameter name and handle all logic

fshabanov (changes requested):

o-nedashkivska:
Seems like you have to update your base branch to main since 40-feat-create-reusable-input-component was merged into main

fshabanov (changes requested):

apps/frontend/src/libs/components/search-input/search-input.tsx
```diff
@@ -0,0 +1,56 @@
...
+	name: FieldPath<T>;
+	onSearch?: (value: string) => void;
+	placeholder?: string;
+};
+
+const SearchInput = <T extends FieldValues>({
+	className = "",
+	control,
+	errors,
+	hasVisuallyHiddenLabel = true,
```
fshabanov, line 29:
```suggestion
	hasVisuallyHiddenLabel = false,
```
Default value should be false

apps/frontend/src/libs/components/search-input/search-input.tsx
```diff
@@ -0,0 +1,56 @@
...
+import styles from "./styles.module.css";
+
+type Properties<T extends FieldValues> = {
+	className?: string;
+	control: Control<T, null>;
+	errors: FieldErrors<T>;
+	hasVisuallyHiddenLabel?: boolean;
+	label: string;
+	name: FieldPath<T>;
+	onSearch?: (value: string) => void;
```
fshabanov, line 21:
```suggestion
	onSearch: (value: string) => void;
```
Should be mandatory, otherwise doesn't make sense to have this component

fshabanov (changes requested):

sofiia-trokhymchuk (approved):

apps/frontend/src/libs/helpers/debounce.helper.ts
```diff
@@ -0,0 +1,22 @@
+type DebounceCallback<T extends unknown[]> = (...arguments_: T) => void;
+
+type DebouncedFunction<T extends unknown[]> = (...arguments_: T) => void;
```
o-nedashkivska, line 3: Do we need two same types here?

apps/frontend/src/libs/hooks/use-app-search/use-app-search.hook.ts
```diff
@@ -0,0 +1,35 @@
...
+	control,
+	debounceDelay = SearchDebounceDelay.DEFAULT,
+	name,
+	onSearch,
+}: UseSearchParameters<T>): void => {
+	const { field } = useFormController({ control, name });
+
+	const watchedValue = field.value as string;
+
+	const debouncedOnSearch = debounce(onSearch, debounceDelay);
```
o-nedashkivska, line 28: If a component that uses this hook is rerendered, does it mean that debouncedOnSearch will be recreated?

fshabanov (approved):
Pull the main branch

PR #54, opened 2025-07-25
feat: get authenticated user ml-12

iLavruk:


apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -1,28 +1,23 @@
...
+
 import {
 	type AuthResponseDto,
+	userService,
 	type UserSignInRequestDto,
 	type UserSignUpRequestDto,
-} from "~/modules/users/users.js";
+} from "../users/users.js";
 
 class AuthService {
```
VolodymyrKozel, line 12: Hey! 👋

Just a quick note about AuthService: to follow SOLID principles (especially Dependency Inversion), it's better to pass UserService into the constructor like this:
```ts
const authService = new AuthService(userService);
```
This makes the code easier to test and change later. Try to avoid creating dependencies inside the class.
Also you have:
```ts
import { encrypt } from "~/libs/modules/encrypt/encrypt.js";
import { jwt } from "~/libs/modules/token/token.js";
```
same issue, so you need:
```ts
const authService = new AuthService(userService, jwt, encrypt);
```

apps/backend/src/modules/auth/auth.ts
```diff
@@ -1,10 +1,9 @@
 import { logger } from "~/libs/modules/logger/logger.js";
-import { userService } from "~/modules/users/users.js";
 
 import { AuthController } from "./auth.controller.js";
 import { AuthService } from "./auth.service.js";
 
-const authService = new AuthService(userService);
+const authService = new AuthService();
```
VolodymyrKozel, line 6: add imports:
```ts
import { encrypt } from "~/libs/modules/encrypt/encrypt.js";
import { jwt } from "~/libs/modules/token/token.js";
```
then
```ts
const authService = new AuthService(userService, encrypt, jwt);
```
hole example
```ts
import { encrypt } from "~/libs/modules/encrypt/encrypt.js";
import { logger } from "~/libs/modules/logger/logger.js";
import { jwt } from "~/libs/modules/token/token.js";
import { userService } from "~/modules/users/users.js";

import { AuthController } from "./auth.controller.js";
import { AuthService } from "./auth.service.js";

const authService = new AuthService(userService, encrypt, jwt);
const authController = new AuthController(logger, authService);

export { authController };
```

iLavruk:
@o-nedashkivska I'm not exactly sure how it happened — I just saw a message saying I should create a new PR, so I went ahead and did that.

fshabanov (approved):

Issue #60, opened 2025-07-26
Add Layout component


o-nedashkivska:
Create a Layout component that includes a Header and a Sidebar. It should also render its children content within the main layout area.
Use this layout inside ProtectedRoute component

PR #71, opened 2025-07-29
feat: add Layout component ml-60

iLavruk:
Implemented Layout component as requested in task #60 to provide consistent page structure with Header and Sidebar for authenticated users.


apps/frontend/src/libs/components/router-provider/protected-route.tsx
```diff
@@ -21,7 +21,7 @@ const ProtectedRoute: React.FC<Properties> = ({
 		);
 	}
 
-	return <>{children}</>;
+	return <Layout>{children}</Layout>;
```
yehor-kostiuk, line 24: you should use it not here but instead of `App` in `index.tsx` in `<ProtectedRoute>`

yehor-kostiuk:
and I also think it's better to do it through `grid-template-areas:` it will be semantically correct, and it will be easier to size the elements and pages 
```
.layout {
	display: grid;
	grid-template-areas:
		"header header"
		"sidebar main";
	grid-template-rows: 74px 1fr;
	grid-template-columns: 263px 1fr;
	height: 100vh;
}
```
and in components
```
.sidebar-container {
	grid-area: sidebar;
```

what do you think @fshabanov

apps/frontend/src/libs/components/layout/layout.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+const Layout: React.FC<Properties> = ({ children }: Properties) => {
+	return (
+		<div className={styles["layout"]}>
+			<Header />
+			<div className={styles["layout__body"]}>
+				<Sidebar>
+					<Navigation items={NAVIGATION_ITEMS} />
+				</Sidebar>
+				<main className={styles["layout__main"]}>{children}</main>
```
yehor-kostiuk, line 18: we can use <Router Outlet> instead of children ? because we have a conflict with protected route, if we leave children , we need to change structure in index.tsx, if we change it to <Router Outlet> see comment above

iLavruk: Got it! I'll implement `RouterOutlet` in `Layout` component instead of children to resolve the routing conflict.

apps/frontend/src/libs/components/router-provider/protected-route.tsx
iLavruk: Got it!

iLavruk:
> and I also think it's better to do it through `grid-template-areas:` it will be semantically correct, and it will be easier to size the elements and pages
> 
> ```
> .layout {
> 	display: grid;
> 	grid-template-areas:
> 		"header header"
> 		"sidebar main";
> 	grid-template-rows: 74px 1fr;
> 	grid-template-columns: 263px 1fr;
> 	height: 100vh;
> }
> ```
> 
> and in components
> 
> ```
> .sidebar-container {
> 	grid-area: sidebar;
> ```
> 
> what do you think @fshabanov

Implemented Grid layout with grid-template-areas as suggested. Thank you for the recommendation.

apps/frontend/src/libs/components/layout/layout.tsx
```diff
@@ -0,0 +1,31 @@
...
+	Sidebar,
+} from "~/libs/components/components.js";
+import { NAVIGATION_ITEMS } from "~/libs/constants/constants.js";
+
+import styles from "./styles.module.css";
+
+const Layout: React.FC = () => {
+	return (
+		<div className={styles["layout"]}>
+			<div className={styles["layout__header"]}>
```
iLavruk, line 14: @fshabanov hi, do I need to create a separate layout__header class here?
Maybe it would be better to: – either move layout__header into the Header component,
– or reuse the existing header class and add grid-area there.

apps/frontend/src/libs/components/layout/layout.tsx
```diff
@@ -0,0 +1,31 @@
...
+				<Header />
+			</div>
+
+			<div className={styles["layout__sidebar"]}>
```
iLavruk, line 18: Same situation as with Header

apps/frontend/src/libs/components/layout/layout.tsx
fshabanov: Since the display: grid is in this file, it would be better to also keep grid-area styles in this file

iLavruk:
> Add screenshots

image

sofiia-trokhymchuk:
update the branch, please, and we are ready to merge 🙌

PR #80, opened 2025-07-31
fix: allow access to Swagger UI ml-72

Stacy-net:
Now Swagger UI can be accessed at `http://localhost:3000/api/v1/documentation` without authentication.

### DIFF
--- a/apps/backend/src/libs/constants/white-routes.constant.ts
+++ b/apps/backend/src/libs/constants/white-routes.constant.ts
@@ -6,7 +6,7 @@ const WHITE_ROUTES: WhiteRoute[] = [
 	{ method: "POST", path: `${APIPath.AUTH}${AuthApiPath.SIGN_UP}` },
 	{ method: "POST", path: `${APIPath.AUTH}${AuthApiPath.SIGN_IN}` },
 
-	{ method: "GET", path: "/documentation*" },
+	{ method: "GET", path: "/documentation" },
 ];
 
 export { WHITE_ROUTES };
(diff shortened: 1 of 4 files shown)


apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -29,6 +29,10 @@ const authorizationPlugin: FastifyPluginCallback<Options> = fp(
 		fastify.addHook(FastifyHook.PRE_HANDLER, async (request) => {
 			const { method, url } = request;
 
+			if (/^\/api\/v\d+\/documentation/.test(url)) {
+				return;
+			}
+
```
sofiia-trokhymchuk, line 35: Why can't the `documentation` endpoint be added to white routes?

Stacy-net: I tried to add all swagger paths to white routes without this part

`if (/^\/api\/v\d+\/documentation/.test(url)) {
				return;
			}`

image

And I've got 401 error:
`{
  "statusCode": 401,
  "error": "Unauthorized",
  "message": "Invalid credentials"
}`

Stacy-net: I seems that I fixed it

apps/backend/src/libs/constants/white-routes.constant.ts
```diff
@@ -5,6 +5,20 @@ import { type WhiteRoute } from "./libs/types/types.js";
 const WHITE_ROUTES: WhiteRoute[] = [
 	{ method: "POST", path: `${APIPath.AUTH}${AuthApiPath.SIGN_UP}` },
 	{ method: "POST", path: `${APIPath.AUTH}${AuthApiPath.SIGN_IN}` },
+
+	{ method: "GET", path: "/documentation" },
```
fshabanov, line 9: Tbh I don't like that you list all routes one by one. Can we do smth like `/documentation/*`?

fshabanov (changes requested):

apps/backend/src/libs/constants/white-routes.constant.ts
fshabanov: Let's use the solution with regex. The current solution is not safe for future possible swagger changes

fshabanov:
Pull the main branch

PR #83, opened 2025-07-31
fix: trim spaces in email before storing ml-74

Roman-PRV:
The problem occurred because the validation schema parsing wasn't executed in the BaseController—which is where trimming should typically take place. I've now added validation to mapHandler. This should resolve the issues with checking and processing all fields before storage.

---

---


apps/backend/src/libs/modules/controller/base-controller.module.ts
```diff
@@ -23,13 +25,29 @@ class BaseController implements Controller {
...
-		request: FastifyRequest,
-		reply: FastifyReply,
-	): Promise<void> {
+	private async mapHandler(options: {
+		handler: APIHandler;
+		reply: FastifyReply;
+		request: FastifyRequest;
+		validation?: ControllerRouteParameters["validation"];
+	}): Promise<void> {
+		const { handler, reply, request, validation } = options;
```
fshabanov, line 34: This should already be done by Fastify when we give validation schema. Try to find other solutions and share with everyone in chat

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -127,14 +134,6 @@ class BaseServerApplication implements ServerApplication {
 		});
 	}
 
-	private initValidationCompiler(): void {
-		this.app.setValidatorCompiler<ValidationSchema>(({ schema }) => {
-			return <T, R = ReturnType<ValidationSchema["parse"]>>(data: T): R => {
-				return schema.parse(data) as R;
```
fshabanov, line 133: I believe the fix should be related to this method. As you can see this method parsed the data and already sets the validatorCompiler. We don't need to add a new one

fshabanov (changes requested):

apps/backend/src/libs/modules/server-application/base-server-application.ts
Roman-PRV: But Fastify does not automatically assign validation results to the request. I did not find any mention in the documentation that it should do this, and in fact, it does not. In the function returned by setValidatorCompiler, we cannot access the request to change anything manually. We can work with data, but the data type does not match the validation result type by default, so typing will break. I don't see how we can do this nicely while being inside the 
```
private initValidationCompiler(): void {
        this.app.setValidatorCompiler<ValidationSchema>(({ schema }) => {
        ...

```

apps/backend/src/libs/modules/controller/base-controller.module.ts
```diff
@@ -23,11 +23,12 @@ class BaseController implements Controller {
...
-		handler: APIHandler,
-		request: FastifyRequest,
-		reply: FastifyReply,
-	): Promise<void> {
+	private async mapHandler(options: {
+		handler: APIHandler;
+		reply: FastifyReply;
+		request: FastifyRequest;
+	}): Promise<void> {
+		const { handler, reply, request } = options;
```
fshabanov, line 31: This change is unnecessary

fshabanov (approved):
We will keep this approach, it looks good

apps/backend/src/libs/modules/controller/base-controller.module.ts
```diff
@@ -53,7 +53,7 @@ class BaseController implements Controller {
 
 		this.routes.push({
 			...options,
-			handler: (request, reply) => this.mapHandler(handler, request, reply),
+			handler: (request, reply) => this.mapHandler(handler, reply, request),
```
fshabanov, line 56: Avoid unnecessary changes

fshabanov (changes requested):

PR #87, opened 2025-07-31
refactor: delete unused code ml-64

yaroslavhroshev:


apps/frontend/src/modules/users/slices/users.slice.ts
```diff
@@ -16,22 +14,11 @@ const initialState: State = {
...
-		});
-		builder.addCase(loadAll.fulfilled, (state, action) => {
-			state.users = action.payload.items;
-			state.dataStatus = DataStatus.FULFILLED;
-		});
-		builder.addCase(loadAll.rejected, (state) => {
-			state.dataStatus = DataStatus.REJECTED;
-		});
-	},
+const { reducer } = createSlice({
```
fshabanov, line 17: I think we can completely remove this slice

apps/frontend/src/modules/users/slices/users.ts
```diff
@@ -1,10 +1 @@
-import { loadAll } from "./actions.js";
-import { actions } from "./users.slice.js";
-
-const allActions = {
-	...actions,
-	loadAll,
-};
-
-export { allActions as actions };
 export { reducer } from "./users.slice.js";
```
fshabanov, line 1: I think we can remove this as well

apps/frontend/src/modules/users/users.ts
```diff
@@ -22,4 +21,4 @@ export {
 	userSignInValidationSchema,
 	userSignUpValidationSchema,
 } from "./libs/validation-schemas/validation-schemas.js";
-export { actions, reducer } from "./slices/users.js";
+export { reducer } from "./slices/users.js";
```
fshabanov, line 24: This too

fshabanov (changes requested):

apps/frontend/src/libs/modules/store/store.module.ts
```diff
@@ -15,7 +15,7 @@ import {
 	authApi,
 	reducer as authReducer,
 } from "~/modules/auth/auth.js";
-import { userApi, reducer as usersReducer } from "~/modules/users/users.js";
+import { userApi } from "~/modules/users/users.js";
```
fshabanov, line 18: Do we still use userApi anywhere?

fshabanov (changes requested):

yaroslavhroshev:
@fshabanov I have a question: do we need this `file apps\backend\src\modules\users\user.controller.ts` and these codes?
Code_B8KliyR5lo
Code_DRpBumURww

I'm not sure for 100% but it seems to me that it's unused code

fshabanov:
@yaroslavhroshev We need to have this in service and repository because our Service and Repository types require that. We can't remove the controller, because we need to use the method from the controller somewhere

fshabanov (approved):

PR #93, opened 2025-08-01
fix: yaml semantic error in meetings.controller.ts ml-86

ulianchenko:
Additionally, I fixed the Fastify warnings 'The body schema is missing' for routes with methods GET, DELETE, and HEAD that do not contain a body.

### DIFF
--- a/apps/backend/src/libs/modules/server-application/base-server-application.ts
+++ b/apps/backend/src/libs/modules/server-application/base-server-application.ts
@@ -140,8 +140,20 @@ class BaseServerApplication implements ServerApplication {
 
 		const schema: Record<string, unknown> = {};
 
-		if (validation?.body && !["DELETE", "GET", "HEAD"].includes(method)) {
+		if (validation?.body) {
 			schema["body"] = validation.body;
+		}
+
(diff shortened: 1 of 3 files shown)


apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -138,12 +138,16 @@ class BaseServerApplication implements ServerApplication {
 	public addRoute(parameters: ServerApplicationRouteParameters): void {
 		const { handler, method, path, validation } = parameters;
 
+		const schema: Record<string, unknown> = {};
+
+		if (validation?.body && !["DELETE", "GET", "HEAD"].includes(method)) {
+			schema["body"] = validation.body;
+		}
```
fshabanov, line 145: Why do we need this?

ulianchenko: I've made the changes, now we check for the presence of validation for each part of the request (body, params, querystring, headers) and add it to the schema only when it exists.

fshabanov (approved):

PR #95, opened 2025-08-02
feat: add new fields to meeting entity ml-29

MrKopchick:


apps/backend/src/db/migrations/20240127205706_add_meetings_table.ts
```diff
@@ -11,6 +11,8 @@ const ColumnName = {
 	HOST: "host",
 	ID: "id",
 	INSTANCE_ID: "instance_id",
+	MEETING_ID: "meeting_id",
```
fshabanov, line 14: You cannot update existing migrations. Create a new one

fshabanov (changes requested):

fshabanov (approved):

PR #111, opened 2025-08-04
fix: add status code if exist ml-96

VolodymyrKozel:


apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -80,7 +81,7 @@ class BaseServerApplication implements ServerApplication {
 							path: issue.path,
 						})),
 						errorType: ServerErrorType.VALIDATION,
-						message: error.message,
+						message: `You have ${String(error.issues.length)} validation issue${error.issues.length === SINGLE_ISSUE_COUNT ? "" : "s"}.`,
```
fshabanov, line 84: What if we want a real validation message?

fshabanov (changes requested):

apps/backend/src/libs/modules/server-application/base-server-application.ts
VolodymyrKozel: The real validation message is JSON-formatted to look good in the console.
```
    "message": "[\n  {\n    \"validation\": \"email\",\n    \"code\": \"invalid_string\",\n    \"message\": \"Email is wrong\",\n    \"path\": [\n      \"email\"\n    ]\n  }\n]"
```
i rollback.

PR #113, opened 2025-08-04
fix: make email validation message more clear ml-100

Karpt221:


o-nedashkivska:
Do we need a separate PR for this? Seems like we already have these changes in https://github.com/BinaryStudioAcademy/bsa-2025-meetlytic/pull/114

PR #116, opened 2025-08-04
fix: return 409 error for duplicate email instead of 500 ml-94

ulianchenko:


apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -48,6 +48,14 @@ class AuthService {
 	public async signUp(
 		userRequestDto: UserSignUpRequestDto,
 	): Promise<AuthResponseDto> {
+		const userWithSameEmail = await this.userService.findByEmail(
+			userRequestDto.email,
+		);
+
+		if (userWithSameEmail) {
+			throw new EmailConflictError();
+		}
```
fshabanov, line 57:
```suggestion
    const hasUserWithSameEmail = Boolean(userWithSameEmail);

		if (userWithSameEmail) {
			throw new AuthError();
		}
```

fshabanov (changes requested):

apps/backend/src/modules/auth/auth.service.ts
ulianchenko: I kept the EmailConflictError() throwing instead of AuthError() so that the error message and code more accurately describe the issue that occurred.

fshabanov: You can pass message and status to AuthError

PR #118, opened 2025-08-04
fix: implement password visibility toggle ml-99

yaroslavhroshev:


apps/frontend/src/libs/enums/input-password.enum.ts
```diff
@@ -0,0 +1,6 @@
+const InputPasswordTypes = {
```
fshabanov, line 1: B23

fshabanov (changes requested):

PR #123, opened 2025-08-05
feat:  display all meetings ml-102

Karpt221:
image

### DIFF
--- a/apps/frontend/src/index.tsx
+++ b/apps/frontend/src/index.tsx
@@ -4,6 +4,7 @@ import { createRoot } from "react-dom/client";
 import "~/assets/css/styles.css";
 import {
 	App,
+	Layout,
 	Link,
 	ProtectedRoute,
 	RouterProvider,
@@ -34,15 +35,20 @@ createRoot(document.querySelector("#root") as HTMLElement).render(
 					{
 						children: [
 							{
-								element: <Meetings />,
-								path: AppRoute.MEETINGS,
+								children: [
+									{
+										element: <Meetings />,
+										path: AppRoute.MEETINGS,
+									},
+								],
+								element: (
+									<ProtectedRoute redirectTo={AppRoute.SIGN_IN}>
+										<Layout />
+									</ProtectedRoute>
+								),
 							},
 						],
-						element: (
-							<ProtectedRoute redirectTo={AppRoute.SIGN_IN}>
-								<App />
-							</ProtectedRoute>
-						),
+						element: <App />,
 					},
 					{
 						element: (
--- a/apps/frontend/src/libs/components/app/app.tsx
+++ b/apps/frontend/src/libs/components/app/app.tsx
@@ -1,4 +1,8 @@
-import { Layout, Loader, SearchInput } from "~/libs/components/components.js";
+import {
+	Loader,
+	RouterOutlet,
+	SearchInput,
+} from "~/libs/components/components.js";
 import { DataStatus } from "~/libs/enums/enums.js";
 import {
 	useAppDispatch,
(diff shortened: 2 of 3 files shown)


apps/frontend/src/index.tsx
```diff
@@ -1,10 +1,10 @@
 import { StrictMode } from "react";
 import { createRoot } from "react-dom/client";
+import { Link } from "react-router-dom";
```
fshabanov, line 3: export from ~/libs/components/components

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -36,18 +32,16 @@ const App: React.FC = () => {
 	return (
 		<>
 			<Loader isLoading={dataStatus === DataStatus.PENDING} withOverlay />
-
-			<div>
-				<RouterOutlet />
+			<Layout />
```
fshabanov, line 35: Why did you remove RouterOutlet from this component?

apps/frontend/src/pages/meetings/components/meeting-item/meeting-item.tsx
```diff
@@ -0,0 +1,31 @@
...
+	date: string;
+	src?: string;
+	title: string;
+};
+
+const MeetingItem: React.FC<Properties> = ({
+	date,
+	src,
+	title,
+}: Properties) => {
```
fshabanov, line 15:
```suggestion
}) => {
```

packages/shared/src/libs/helpers/format-date/format-date.ts
```diff
@@ -0,0 +1,7 @@
+import dayjs from "dayjs";
+
+const formatDate = (date: Date, format: string): string => {
```
fshabanov, line 3: Create explicit format type

fshabanov (changes requested):

apps/frontend/src/libs/components/app/app.tsx
Karpt221: I removed RouterOutlet from App because Layout, which declared inside the App, already has RouterOutlet.

```typescript
const Layout: React.FC = () => {
	return (
		<div className={styles["layout"]}>
			<div className={styles["layout__header"]}>
				<Header />
			</div>

			<div className={styles["layout__sidebar"]}>
				<Sidebar>
					<Navigation items={NAVIGATION_ITEMS} />
				</Sidebar>
			</div>

			<main className={styles["layout__main"]}>
				<RouterOutlet />
			</main>
		</div>
	);
};

``` 

Would it be better to move Layout to the react router to be able to define differnet app layouts?
```typescript
{
						children: [
							{
								element: <Layout />,
								children: [{
								    element: <Meetings />,
								    path: AppRoute.MEETINGS,
								}]
							},
						],
						element: (
							<ProtectedRoute redirectTo={AppRoute.SIGN_IN}>
								<App />
							</ProtectedRoute>
						),
					},
```

apps/frontend/src/pages/meetings/components/meeting-item/meeting-item.tsx
Karpt221: If I remove Properties Ieslint will throw:
'date'(e.t.c.) is missing in props validationeslint[react/prop-types]

packages/shared/src/libs/helpers/format-date/format-date.ts
Karpt221: Should it be a union type?

type DateFormat = "MMM D, YYYY, h:m A" | "YYYY-MM-DD" e.t.c

fshabanov: Yes

packages/shared/src/index.ts
```diff
@@ -22,6 +22,7 @@ export {
 } from "./libs/modules/http/http.js";
 export { type Storage } from "./libs/modules/storage/storage.js";
 export {
+	type DateFormat,
```
fshabanov, line 25: I don't think we need to export it. It will be only inside formatDate helper

fshabanov (changes requested):

fshabanov (approved):

PR #124, opened 2025-08-05
refactor(shared): convert HTTP methods type to enum ml-61

VolodymyrKozel:


packages/shared/tsconfig.json
```diff
@@ -3,6 +3,9 @@
 	"include": ["**/*.ts", "**/*.js"],
 	"exclude": ["node_modules", "build"],
 	"compilerOptions": {
-		"baseUrl": "."
+		"baseUrl": ".",
+		"paths": {
+			"~/*": ["./src/*"]
+		}
```
fshabanov, line 9: We don't need this in shared. In shared we will use relative path

fshabanov (changes requested):

fshabanov (approved):

PR #163, opened 2025-08-10
feat: add meeting public url ml-143

Karpt221:


apps/backend/src/modules/meetings/meetings.controller.ts
```diff
@@ -121,6 +122,11 @@ class MeetingsController extends BaseController {
 			path: MeetingsApiPath.ROOT,
 			preHandlers: [checkIfMeetingOwner(this.meetingService)],
 		});
+		this.addRoute({
+			handler: (options) => this.getPublicUrl(options as GetPublicUrlOptions),
+			method: "GET",
```
o-nedashkivska, line 127: We have enum for methods

packages/shared/src/modules/meetings/libs/types/meeting-detailed-response-dto.type.ts
```diff
@@ -0,0 +1,16 @@
+import { type ValueOf } from "../../../../libs/types/types.js";
+import { type MeetingHost, type MeetingStatus } from "../enums/enums.js";
+
+type MeetingDetailedResponseDto = {
+	actionItems: null | string;
+	createdAt: string;
+	host: ValueOf<typeof MeetingHost>;
+	id: number;
+	instanceId: null | string;
```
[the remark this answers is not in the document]
Karpt221: We need to retrieve the instanceId from the entity in order to update the meeting. At the same time, the client should not receive the instanceId. Can I create a getClientObject() method that returns the same data as toObject() but omits the instanceId?


```typescript
	public async update(
		id: number,
		payload: MeetingUpdateRequestDto,
	): Promise<MeetingResponseDto> {
		const meetingEntity = await this.meetingRepository.find(id);

		if (!meetingEntity) {
			throw new MeetingError({
				message: MeetingErrorMessage.CANNOT_UPDATE_NON_EXISTENT,
				status: HTTPCode.NOT_FOUND,
			});
		}

		const meeting = MeetingEntity.initialize({
			actionItems: meetingEntity.toDetailedObject().actionItems,
			createdAt: meetingEntity.toObject().createdAt,
			host: payload.host,
			id,
			instanceId: meetingEntity.toObject().instanceId,
			meetingId: meetingEntity.toObject().meetingId,
			meetingPassword: meetingEntity.toObject().meetingPassword,
			ownerId: meetingEntity.toObject().ownerId,
			status: payload.status,
			summary: meetingEntity.toDetailedObject().summary,
		});

		const updatedMeeting = await this.meetingRepository.update(
			id,
			meeting.toNewObject(),
		);

		if (!updatedMeeting) {
			throw new MeetingError({
				message: MeetingErrorMessage.UPDATE_FAILED,
				status: HTTPCode.BAD_REQUEST,
			});
		}

		return updatedMeeting.toObject(); // toClentObject()
	}

```

apps/backend/src/modules/meetings/meetings.service.ts
```diff
@@ -132,6 +147,18 @@ class MeetingService implements Service<MeetingResponseDto> {
 		return { items: meetings.map((meeting) => meeting.toObject()) };
 	}
 
+	public async getPublicUrl(
+		id: number,
+	): Promise<MeetingGetPublicUrlResponseDto> {
+		await this.find(id);
```
fshabanov, line 153: Why do we need this find if we don't use it?

packages/shared/src/modules/meetings/libs/types/meeting-detailed-response-dto.type.ts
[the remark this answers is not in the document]
fshabanov: Yes, you can

apps/backend/src/modules/meetings/meetings.service.ts
Karpt221: "I thought that it would allow us to reuse the code."


```typescript

public async find(id: number): Promise<MeetingDetailedResponseDto> {

const meeting = await this.meetingRepository.find(id);

if (!meeting) {
throw new MeetingError({
message: MeetingErrorMessage.MEETING_NOT_FOUND,
status: HTTPCode.NOT_FOUND,
});

}

return meeting.toDetailedObject();
}

public async getPublicUrl(
id: number,
): Promise<MeetingGetPublicUrlResponseDto> {
await this.find(id);

/*
instead of

const meeting = await this.meetingRepository.find(id);
          
if (!meeting) {
 throw new MeetingError({
    message: MeetingErrorMessage.MEETING_NOT_FOUND,
    status: HTTPCode.NOT_FOUND,
});
} */

const token = await this.sharedJwt.sign({ meetingId: id });

return {
publicUrl: `${APIPath.PUBLIC_MEETINGS}/${String(id)}?token=${token}`,
};

}

```

fshabanov: find method will always return a meeting, so you don't need to check if the meeting doesn't exist. Let's do this way so that it makes more sendse:
```
const meeting = await this.find(id);

const token = await this.sharedJwt.sign({meetingId: meeting.id});
```

apps/backend/src/modules/meetings/meetings.controller.ts
```diff
@@ -280,6 +333,46 @@ class MeetingsController extends BaseController {
...
+	 *         content:
+	 *           application/json:
+	 *             schema:
+	 *               $ref: "#/components/schemas/Meeting"
+	 *       404:
+	 *         description: Meeting not found
+	 *       401:
+	 *         description: Signed URL was malformed
+	 */
+	private async verifyUrl(
```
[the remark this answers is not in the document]
fshabanov: Yes, you're right. Should be as a query param. In that case let's use `/meetings/:id/public?token={token}`
And I agree with you that we need to rename it

apps/backend/src/modules/meetings/meetings.controller.ts
```diff
@@ -243,6 +260,80 @@ class MeetingsController extends BaseController {
...
+	 *         description: Meeting not found
+	 *       401:
+	 *         description: Signed URL was malformed
+	 */
+	private async findBySignedUrl(
+		options: FindBySignedUrlOptions,
+	): Promise<APIHandlerResponse> {
+		const id = Number(options.params.id);
+		const token = options.query.token;
+		const url = await this.meetingService.findBySignedUrl(id, token);
```
fshabanov, line 298: Why url?

apps/backend/src/modules/meetings/meetings.controller.ts
```diff
@@ -243,6 +260,80 @@ class MeetingsController extends BaseController {
 		return { payload: meetings, status: HTTPCode.OK };
 	}
 
+	/**
+	 * @swagger
+	 * /meetings/{id}/token:
```
fshabanov, line 265:
```suggestion
	 * /meetings/{id}/public:
```

ulianchenko (approved):

PR #167, opened 2025-08-11
fix: generate different error codes for validation signup and signin ml-162

Roman-PRV:
- [x] The API should respond with 422 Unprocessable Entity for validation errors related to the format and content of signup and signin fields.
- [x] It should respond with 400 Bad Request only when the JSON payload is malformed ~~or missing required keys (e.g., email or password key missing)~~.

image

---

image

---

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -51,7 +51,7 @@
 		},
 		"apps/backend": {
 			"name": "@meetlytic/backend",
-			"version": "1.4.0",
+			"version": "1.4.1",
 			"dependencies": {
 				"@aws-sdk/client-cloudformation": "3.848.0",
 				"@fastify/static": "8.2.0",
@@ -99,7 +99,7 @@
 		},
 		"apps/frontend": {
 			"name": "@meetlytic/frontend",
-			"version": "1.4.0",
+			"version": "1.4.1",
 			"dependencies": {
 				"@hookform/resolvers": "4.1.3",
 				"@meetlytic/shared": "*",
@@ -15588,7 +15588,7 @@
 		},
 		"packages/shared": {
 			"name": "@meetlytic/shared",
-			"version": "1.3.0",
+			"version": "1.3.1",
 			"dependencies": {
 				"dayjs": "1.11.13"
 			},


apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -67,30 +68,41 @@ class BaseServerApplication implements ServerApplication {
...
-						this.logger.error(`[${issue.path.toString()}] — ${issue.message}`);
+					for (let validation of error.validation) {
+						this.logger.error(
+							`[${validation.params.issue.path.toString()}] — ${validation.message}`,
+						);
 					}
 
+					const missingKey = error.validation.some(
+						(validation) => validation.message === "Required",
+					);
```
fshabanov, line 90: We shouldn't handle status based on validation message. Status is defined when throwing an error

fshabanov (changes requested):

apps/backend/src/libs/modules/server-application/base-server-application.ts
Roman-PRV: @fshabanov 
The validation error thrown by Fastify+Zod always has a status of 400. If we want to return 422, we have to set this status manually. And if we want to have different statuses for different validation errors, in particular 400 for .required(), we need to somehow distinguish between these errors. Through “message”: “Required”, or through “keyword”: “invalid_type” or through ‘code’: “invalid_type”.
Other options for solving this problem — creating a custom validatorCompiler or adding custom errors to the validation scheme — seem to me to be much more cumbersome and will lead to increased coupling between system components.

sofiia-trokhymchuk: @fshabanov I agree that for such case switching between statuses based on the error message is fine. but there is no need to replace `error.issues` with `error.validation.` use `issue.code` to assign appropriate HTTP status: e.g. "invalid_type" and "unrecognized_keys" would return 422
https://zod.dev/packages/core#errors

Roman-PRV: @fshabanov @sofiia-trokhymchuk 
> There should be a way to set a default status for Zod errors. All validation errors should have 422 by default, we don't want to manually handle every single one

Yes, we can set 422 for all Zod errors, which is what we do. But the absence of a field that is checked as `.required` in the schema also refers to Zod errors, so it will also be 422. If we want to fulfill the condition “It should respond with 400 Bad Request only when ...or missing required keys (e.g., email or password key missing),” we have to handle them manually.

So, what do we do? Do we handle missing fields manually via `issue.code` to give them a 400 code or do we return 422 for all validation errors  (including missing fields)?

Roman-PRV: So, according to https://binarystudioacademyhq.slack.com/archives/C095TFFFBRV/p1754988061229429?thread_ts=1754986627.445199&cid=C095TFFFBRV
 all errors Zod will be 422.

fshabanov (approved):

fshabanov (approved):

PR #184, opened 2025-08-13
feat: end meeting recording from the frontend ml-160

Karpt221:
I thought about "Stop recording" flow for a while. It seems that "Stop recording" feature tightly coupled with bot lifecycle and websockets. Since we don't have websockets yet and bot lifecycle is not implemented either, I will implement this feature partially. The current implementation only accepts a request, sets meeting status as "ended" in the db and deletes a stack. In the future we need to refactor our implementation  of this feature, because stop recordong endpoint should only emit a message to bot. Stack should only be deleted after bot has sent message to the backend.

Stop Recording flow

1. User clicks "Stopt Recording" button and sends request to backend.

2. When the backend receives a request we emit a message for the bot. (websockets are not implemented)

3. Bot finalizes everything (stops current chunk recording, stops recording of full audio, transcribes full audio, gets summary and action points), then emits event to Backend. (bot lifecycle is not implemented)

4. At this point Backend deletes CloudFormation stack, saves  transcription text, summary, action items  to db and sends them the frontend using websockets.


fshabanov (approved):
Backend seems good, waiting for frontend

fshabanov (approved):

yaroslavhroshev (approved):

PR #204, opened 2025-08-15
fix: no internet toast notification ml-173

yaroslavhroshev:


apps/frontend/src/libs/helpers/extract-error-message.helper.ts
```diff
@@ -1,23 +1,53 @@
...
-			"data" in maybeResponse &&
-			typeof (maybeResponse as { data?: unknown }).data === "object"
-		) {
-			const maybeMessage = (maybeResponse as { data: { message?: unknown } })
-				.data.message;
-
-			if (typeof maybeMessage === "string") {
-				return maybeMessage;
-			}
+	if (isCustomHttpError(error)) {
```
fshabanov, line 29: No need for all that stuff. Before sending a request check if the user is online. If not, throw an error with a custom message

fshabanov (changes requested):

PR #212, opened 2025-08-16
fix: bot unable to join zoom meetings requiring auth user ml-205

MrKopchick:


fshabanov (commented):
Proceed with testing and let us know about the results

PR #226, opened 2025-08-18
feat: feat bot lifecycle ml-161

Karpt221:
image

**Summary:**

-   "Stop recording" feature was implemented using Websockets message chain
-  Two websocket namespaces were added for bots and front-end users to keep their rooms separate but with the same names.
- After clicking "Stop recording" button, bot leaves a meeting and sends generated summary / action items to the backend, backend saves them in the database. After that backend deletes meeting stack.

image

PR #233, opened 2025-08-18
fix: allow reconnection zoom meeting ml-230

Roman-PRV:
- [ ] ~~If the bot is disconnected out of the meeting, the meeting status should automatically be marked as "ended".~~ This component is implemented in feat: feat bot lifecycle ml-161 #226
- [x] The bot should be able to rejoin the meeting using the same link without errors, as long as no active bot is currently connected
- [x] Recurring meetings should be handled in the same way, allowing the bot to reconnect when the meeting has been marked as ended

---

- [x] remove unique constraint on db level
- [x] If there is a meeting with the same link/id and its status is "started" - don't allow creating a meeting
- [x] If the status is "ended" or there's no meeting - allow creating a meeting

---

image


packages/shared/src/libs/constants/constants.ts
```diff
@@ -1,3 +1,4 @@
+export { EMPTY_ARRAY_LENGTH } from "./empty-array-lenght.js";
```
fshabanov, line 1:
```suggestion
export { EMPTY_ARRAY_LENGTH } from "./empty-array-length.js";
```

fshabanov (approved):

VolodymyrKozel (approved):

fshabanov (approved):

PR #242, opened 2025-08-19
fix: quality criteria cycle 1 ml-237

yehor-kostiuk:


apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -83,6 +83,7 @@ const Input = <T extends FieldValues>({
 						{isClickableIcon ? (
 							<button onClick={onClickIcon} type="button">
 								{icon}
+								<span className="visually-hidden">{iconName}</span>
```
fshabanov, line 86: iconName is not very descriptive, maybe something else?

apps/frontend/src/modules/users/libs/default-values/sign-up.default-values.ts
```diff
@@ -1,11 +1,11 @@
 import { type UserSignUpRequestDto } from "~/modules/users/users.js";
 
-const DEFAULT_SIGN_UP_PAYLOAD: UserSignUpRequestDto = {
+const DEFAULT_SIGN_UP_VALUES: UserSignUpRequestDto = {
```
fshabanov, line 3:
```suggestion
const USER_SIGN_UP_DEFAULT_VALUES: UserSignUpRequestDto = {
```
Also change file name to `user-sign-up-default-values.constant.ts`. Also change the folder name to `constants`

packages/shared/src/libs/exceptions/exceptions.ts
```diff
@@ -1,6 +1,6 @@
-export { AuthError } from "./auth-error/auth.error.js";
+export { AuthError } from "./auth-error/auth-error.exeption.js";
```
fshabanov, line 1:
```suggestion
export { AuthError } from "./auth-error/auth-error.exception.js";
```

fshabanov (changes requested):

PR #252, opened 2025-08-20
fix: copy public meeting link ml-244

yehor-kostiuk:


apps/frontend/src/pages/meeting-details/meeting-details.tsx
```diff
@@ -97,7 +97,16 @@ const MeetingDetails: React.FC = () => {
 					meeting.id,
 				);
 				const host = config.ENV.APP.HOST;
-				void navigator.clipboard.writeText(`${host}${publicUrl}`);
+				const fullUrl = `${host}${publicUrl}`;
+
+				const textarea = document.createElement("textarea");
+				textarea.value = fullUrl;
+				document.body.append(textarea);
+				textarea.select();
```
fshabanov, line 105: Just in case, move textarea outside of visible view and make its opacity 0. I know we remove it, but just in case

fshabanov (approved):

apps/frontend/src/pages/meeting-details/meeting-details.tsx
fshabanov: Yes, sure. Smth like `textarea.style.opacity = 0`. Same for moving outside of view

fshabanov (approved):

PR #296, opened 2025-08-24
fix: changed rings animation and add swipe for testimonials on mobile ml-293

iLavruk:


apps/frontend/src/modules/landing/libs/helpers/rings.helper.ts
```diff
@@ -36,15 +36,40 @@ const initRings = ({
...
+
+	const scaledRingSize = isMobile
+		? {
+				[LandingBgRingType.LARGE]:
+					LandingBgRingSize.LARGE / LandingBgInit.MOBILE_RING_SIZE_DIVISOR,
+				[LandingBgRingType.MIDDLE]:
+					LandingBgRingSize.MIDDLE / LandingBgInit.MOBILE_RING_SIZE_DIVISOR,
+				[LandingBgRingType.SMALL]:
+					LandingBgRingSize.SMALL / LandingBgInit.MOBILE_RING_SIZE_DIVISOR,
+			}
```
o-nedashkivska, line 50: These values are static (meaning they do not depend on `initRings` inputs), so let's put them to other enums
Btw, not sure if you need MOBILE_RING_SIZE_DIVISOR, you can just define specific values for mobile ring sizes the same as you did with LandingBgRingSize

apps/frontend/src/modules/landing/libs/helpers/rings.helper.ts
```diff
@@ -36,15 +36,40 @@ const initRings = ({
...
+		: LandingBgRingSize;
+
+	const MOBILE_SPEED_SCALE_MULTIPLIER = 1.5;
```
o-nedashkivska, line 53: Should be moved to constants

apps/frontend/src/modules/landing/libs/helpers/rings.helper.ts
```diff
@@ -36,15 +36,40 @@ const initRings = ({
...
+
+	const scaledSpeedScale = isMobile
+		? {
+				[LandingBgRingType.LARGE]:
+					LandingBgRingSpeedScale.LARGE * MOBILE_SPEED_SCALE_MULTIPLIER,
+				[LandingBgRingType.MIDDLE]:
+					LandingBgRingSpeedScale.MIDDLE * MOBILE_SPEED_SCALE_MULTIPLIER,
+				[LandingBgRingType.SMALL]:
+					LandingBgRingSpeedScale.SMALL * MOBILE_SPEED_SCALE_MULTIPLIER,
+			}
```
o-nedashkivska, line 63: The same as for `scaledRingSize`

fshabanov (approved):

PR #297, opened 2025-08-24
fix: handle network errors ml-173

iLavruk:


apps/frontend/src/libs/helpers/is-network-error.helper.ts
```diff
@@ -0,0 +1,9 @@
+const isNetworkError = (error: unknown): boolean => {
+	if (typeof error === "object" && error !== null) {
+		return (error as { code?: string }).code === "ERR_NETWORK";
```
fshabanov, line 3:
```suggestion
		return (error as { code?: string }).code === NETWORK_ERROR_CODE;
```

fshabanov (approved):

iLavruk:
This task is no longer relevant.

PR #316, opened 2025-08-25
fix: add 405 response for unsupported methods ml-182

Roman-PRV:
- [x] the server returns a 405 Method Not Allowed status code when an unsupported HTTP method is used the application's endpoints.

image

---

image


apps/backend/src/libs/plugins/method-guard/method-guard.plugin.ts
```diff
@@ -0,0 +1,56 @@
...
+		const { found, methodAllowed } = matchRoute(url, method, allRoutes);
+
+		if (!found) {
+			return response.status(HTTPCode.NOT_FOUND).send({
+				errorType: ServerErrorType.COMMON,
+				message: ExceptionMessage.ROUTE_NOT_FOUND,
+			});
+		}
+
+		if (!methodAllowed) {
```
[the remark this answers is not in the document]
Roman-PRV: @fshabanov 
I tried doing this in errrorHandler. It worked for authorized users, but the authorization plugin throws a 401 error for unauthorized users before fastify has time to process 405.

apps/backend/src/libs/plugins/method-guard/libs/match-route.ts
```diff
@@ -0,0 +1,22 @@
...
+	url: string,
+	method: string,
+	allRoutes: { method: string; path: string }[],
+): { found: boolean; methodAllowed: boolean } {
+	const matchingRoutes = allRoutes.filter((route) =>
+		match(route.path, { decode: decodeURIComponent })(url),
+	);
+
+	const found = matchingRoutes.length > EMPTY_ARRAY_LENGTH;
+	const methodAllowed = matchingRoutes.some(
```
fshabanov, line 15:
```suggestion
	const isRouteFound = matchingRoutes.length > EMPTY_ARRAY_LENGTH;
	const isMethodAllowed = matchingRoutes.some(
```

apps/backend/src/libs/plugins/method-guard/method-guard.plugin.ts
```diff
@@ -0,0 +1,56 @@
...
+type Options = {
+	allRoutes: { method: string; path: string }[];
+};
+
+const methodGuardCallback: FastifyPluginCallback<Options> = (
+	fastify,
+	options,
+	done,
+) => {
+	fastify.addHook("onRequest", (request, response, next) => {
```
fshabanov, line 18: We have FastifyHook enum

apps/backend/src/libs/plugins/method-guard/method-guard.plugin.ts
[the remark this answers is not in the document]
fshabanov: You're right, fastify doesn't handle this

apps/backend/src/libs/plugins/method-guard/method-guard.plugin.ts
```diff
@@ -0,0 +1,56 @@
...
+		const { allRoutes } = options;
+		const method = request.method.toUpperCase();
+		const [url] = response.request.url.split("?");
+
+		if (!url) {
+			return response.status(HTTPCode.NOT_FOUND).send({
+				errorType: ServerErrorType.COMMON,
+				message: ExceptionMessage.ROUTE_NOT_FOUND,
+			});
+		}
```
fshabanov, line 28: Can we throw an error instead? Then the error will caught by error handler

apps/backend/src/libs/plugins/method-guard/method-guard.plugin.ts
```diff
@@ -0,0 +1,56 @@
...
+
+		const { found, methodAllowed } = matchRoute(url, method, allRoutes);
+
+		if (!found) {
+			return response.status(HTTPCode.NOT_FOUND).send({
+				errorType: ServerErrorType.COMMON,
+				message: ExceptionMessage.ROUTE_NOT_FOUND,
+			});
+		}
```
fshabanov, line 37: Same here

apps/backend/src/libs/plugins/method-guard/method-guard.plugin.ts
```diff
@@ -0,0 +1,56 @@
...
+
+		if (!methodAllowed) {
+			return response.status(HTTPCode.METHOD_NOT_ALLOWED).send({
+				errorType: ServerErrorType.COMMON,
+				message: ExceptionMessage.METHOD_NOT_ALLOWED,
+			});
+		}
```
fshabanov, line 44: Same here, throw an error

fshabanov (changes requested):

PR #328, opened 2025-08-26
fix: convert markdown to styles for pdf export ml-301

Roman-PRV:
- [x] Markdown is rendered into proper PDF formatting

image

---

image


apps/frontend/src/libs/components/meeting-pdf/libs/components/render-markdown.tsx
```diff
@@ -0,0 +1,43 @@
...
+	return (
+		<View>
+			{lines.map((line, index) => {
+				const key = `line-${index.toString()}`;
+
+				if (!line.trim()) {
+					return <View key={key} style={styles.paragraph} />;
+				}
+
+				const astLine = parseMarkdown(line);
```
fshabanov, line 26: What is ast?

fshabanov (changes requested):

apps/frontend/src/libs/components/meeting-pdf/libs/components/render-markdown.tsx
fshabanov: Rename, because the naming is not clear

apps/frontend/src/libs/components/meeting-pdf/libs/components/render-node.tsx
```diff
@@ -0,0 +1,85 @@
...
+	type BlockContent,
+	type ListItem,
+	type PhrasingContent,
+} from "~/libs/types/types.js";
+
+import { styles } from "../helpers/meeting-pdf.styles.js";
+
+type MdNode = BlockContent | ListItem | PhrasingContent;
+
+function renderNode(node: MdNode, key: string): React.ReactElement {
```
fshabanov, line 13: Use arrow functions

apps/frontend/src/libs/components/meeting-pdf/libs/components/render-node.tsx
```diff
@@ -0,0 +1,85 @@
+import { Text, View } from "~/libs/components/components.js";
+import { HEADING_LEVEL_PRIMARY } from "~/libs/constants/constants.js";
+import {
+	type BlockContent,
+	type ListItem,
+	type PhrasingContent,
+} from "~/libs/types/types.js";
+
+import { styles } from "../helpers/meeting-pdf.styles.js";
```
fshabanov, line 9: styles is not a helper. Helpers are functions

apps/frontend/src/libs/components/meeting-pdf/meeting-pdf.tsx
```diff
@@ -1,42 +1,38 @@
+import React from "react";
```
fshabanov, line 1: We don't need to import React

fshabanov (changes requested):

fshabanov (approved):

PR #335, opened 2025-08-27
fix: no automatic logout and redirect when token invalid ml-174

yehor-kostiuk:


apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -4,8 +4,10 @@ import {
 	useAppDispatch,
 	useAppSelector,
 	useCallback,
+	useEffect,
 	useLocation,
 } from "~/libs/hooks/hooks.js";
+import { notification } from "~/libs/modules/notifications/notifications.js";
```
fshabanov, line 10: We shouldn't import modules directly to components

yehor-kostiuk: we have a notification in the meetings details component, I added this message because some simple actions on the page do not trigger an action and do not return 401 and there will be no message as a result, so I tried to cover all possible situations with this

PR #336, opened 2025-08-27
feat: export button should be disabled during live transcription ml-330

MrKopchick:


apps/frontend/src/pages/meeting-details/meeting-details.tsx
```diff
@@ -46,6 +46,7 @@ import { actions as transcriptionActions } from "~/modules/transcription/transcr
 import styles from "./styles.module.css";
 
 const MeetingDetails: React.FC = () => {
+	const EMPTY = 0;
```
fshabanov, line 49: We have EMPTY_ARRAY_LENGTH already

PR #338, opened 2025-08-27
feat: use consistent datetime format ml-308

Karpt221:
image

image

image


packages/shared/src/libs/enums/datetime-format.enum.ts
```diff
@@ -0,0 +1,6 @@
+const DateTimeFormat = {
+	AUDIO_PLAYER: "mm:ss",
+	MEETING: "MMM D, YYYY, h:mm A",
+} as const;
```
fshabanov, line 4: I don't think we need this enum

VolodymyrKozel (approved):

PR #375, opened 2025-08-28
fix: signup-mobile-firefpx ml-374

iLavruk:
image


fshabanov (approved):
Did you test with chrome as well?

Issue #351, opened 2025-08-28
fix: Public URL should Look Identical to the Meeting Details Page of a Logged in User


TheGiodude:
### Describe your issue

**After our Product Owner Nikita's request, our application looked like this:**

<img width="2559" height="1273" alt="Image" src="https://github.com/user-attachments/assets/fce4aeb6-cecf-4deb-acb0-e42c8439979b" />

**And the meeting details page looked like this:**

<img width="2083" height="1205" alt="Image" src="https://github.com/user-attachments/assets/34f43a72-eb19-40a8-868e-43e2e6179aa3" />

**The public URL of that same page looks like this:**

<img width="2559" height="1269" alt="Image" src="https://github.com/user-attachments/assets/cc7793e2-ea2a-45e0-becd-20c0c9f6e849" />

### Steps to reproduce

1. Go to a meeting's details page
2. Observe the view
3. Go to the Public URL of that same page
4. Compare the views

### What was the actual result?

The Public URL page is not adjusted in the same way as the rest of the application

### What was the expected result?

The Public URL page (even though it lacks certain elements) should look exactly like the meeting details page of a logged in user

### What OS are you seeing the problem on?

Windows

### What browsers are you seeing the problem on?

Firefox

### Severity

Minor (S4)

PR #385, opened 2025-08-29
fix: limit the width of public layout ml-351

Roman-PRV:
- [x] fix: Public URL should Look Identical to the Meeting Details Page of a Logged in User #351

image


Roman-PRV:
wrong branch name

